PR #43, opened 2016-02-08
Test fetch collection

skomputer:


src/custom-typings/jest.d.ts
```diff
@@ -0,0 +1,30 @@
+// Compiled using typings@0.6.6
+// Source: https://raw.githubusercontent.com/DefinitelyTyped/DefinitelyTyped/86dbea8fc37d9473fee465da4f0a21bea4f8cbd9/jest/jest.d.ts
+// Type definitions for Jest 0.1.18
+// Project: http://facebook.github.io/jest/
+// Definitions by: Asana <https://asana.com>
+// Definitions: https://github.com/borisyankov/DefinitelyTyped
```
aslagle: we need to take out this comment and explain why we wrote this file instead of using those typings

typings.json
```diff
@@ -1,6 +1,6 @@
 {
   "ambientDependencies": {
-    "jest": "github:DefinitelyTyped/DefinitelyTyped/jest/jest.d.ts#86dbea8fc37d9473fee465da4f0a21bea4f8cbd9",
+    "nock": "github:DefinitelyTyped/DefinitelyTyped/nock/nock.d.ts#d14a6d94a642db133d6aeaca64f075033bba12c9",
```
aslagle: did you end up using nock?

src/actions.ts
```diff
@@ -1,21 +1,28 @@
...
-    dispatch(requestCollection(url));
-    console.log("fetching url: ", url)
-
-    return fetchOPDSData(url, (data) => {
-      dispatch(loadCollection(data, url));
+    dispatch(fetchCollectionRequest(url));
+    return new Promise((resolve, reject) => {
+      fetchOPDSData(url).then((data: CollectionProps) => {
+        dispatch(fetchCollectionSuccess());
+        dispatch(loadCollection(data, url));
```
aslagle: I don't remember if you said why you were splitting up these actions, but now I'm not sure it makes sense. We wouldn't want the loading message to go away before the data is there. Maybe they should be separate but we should check a different property to decide whether to show the loading indicator.

src/actions.ts
```diff
@@ -1,21 +1,28 @@
...
-
-    return fetchOPDSData(url, (data) => {
-      dispatch(loadCollection(data, url));
+    dispatch(fetchCollectionRequest(url));
+    return new Promise((resolve, reject) => {
+      fetchOPDSData(url).then((data: CollectionProps) => {
+        dispatch(loadCollection(data, url));
+        dispatch(fetchCollectionSuccess());
+        resolve(data);
+      }).catch(err => reject(err));
```
aslagle: Maybe this should dispatch a fetch collection failure action instead of rejecting the promise. I'm not sure anything will handle the rejection further up...

PR #46, opened 2016-02-09
Routing

### LINKED ISSUE #29: Set up routing
Bookmarkable/shareable URLs for each view

skomputer:
This will close #29.

### DIFF
--- a/src/components/CollectionLink.tsx
+++ b/src/components/CollectionLink.tsx
@@ -1,24 +1,20 @@
 import * as React from "react";
 
-export default class CollectionLink extends React.Component<any, any> {
-  constructor(props: CollectionLinkProps) {
-    super(props);
-  }
-
+export default class CollectionLink extends React.Component<CollectionLinkProps, any> {
   render(): JSX.Element {
     return (
       <a className={this.props.className} href={"?url=" + this.props.url} onClick={this.handleClick.bind(this)}>
-        {this.props.text || this.props.children}
+        {this.props.text}
       </a>
     );
   }
 
   handleClick(event) {
     // if any of these keys are pressed, the url is opened by the browser as it pleases
-    if (event.metaKey || event.altKey || event.ctrlKey) {
+    if (event.metaKey || event.altKey || event.ctrlKey || event.shiftKey) {
       return true;
     } else {
-      // if not, the provided url is passed to the provided onClick, if url and onClick exist
+      // if not, the url is passed to fetchCollection
       this.props.fetchCollection(this.props.url);
       event.preventDefault();
       return false;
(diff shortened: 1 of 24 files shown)


src/components/CollectionLink.tsx
```diff
@@ -0,0 +1,27 @@
+import * as React from "react";
+
+export default class CollectionLink extends React.Component<any, any> {
```
aslagle: should this be ```React.Component<CollectionLinkProps, any>```?

src/components/CollectionLink.tsx
```diff
@@ -0,0 +1,27 @@
...
+  constructor(props: CollectionLinkProps) {
+    super(props);
+  }
```
aslagle: do you need the constructor?

src/components/CollectionLink.tsx
```diff
@@ -0,0 +1,27 @@
...
+    return (
+      <a className={this.props.className} href={"?url=" + this.props.url} onClick={this.handleClick.bind(this)}>
+        {this.props.text || this.props.children}
+      </a>
+    );
+  }
+
+  handleClick(event) {
+    // if any of these keys are pressed, the url is opened by the browser as it pleases
+    if (event.metaKey || event.altKey || event.ctrlKey) {
```
aslagle: :+1: and let's also add shift just in case. In chrome, shift click opens a new window instead of a new tab.

src/components/Root.tsx
```diff
@@ -45,14 +55,34 @@ const mapStateToProps = (state) => {
...
     fetchSearchDescription: (url) => dispatch(fetchSearchDescription(url))
   }
 }
 
+const mergeProps = (stateProps, dispatchProps, ownProps) => {
+  return Object.assign({}, ownProps, stateProps, dispatchProps, {
+    fetchCollection: (url: string, useHandler: boolean = true) => {
+      dispatchProps.fetchCollection(url);
+
+      if (useHandler && ownProps.onFetch) {
```
aslagle: let's rename useHandler to something like skipOnFetch

aslagle: that would also change the default to false

src/components/CollectionLink.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+export default class CollectionLink extends React.Component<any, any> {
+  constructor(props: CollectionLinkProps) {
+    super(props);
+  }
+
+  render(): JSX.Element {
+    return (
+      <a className={this.props.className} href={"?url=" + this.props.url} onClick={this.handleClick.bind(this)}>
+        {this.props.text || this.props.children}
```
aslagle: maybe we shouldn't have the children option if we're not using it?

aslagle:
:+1:

PR #59, opened 2016-02-11
Fix parser typings

aslagle:
Updated parser version and added typings that are no longer published with the parser.

This connects to https://github.com/NYPL-Simplified/opds-feed-parser/issues/29


skomputer:
:+1:

PR #63, opened 2016-02-11
Scroll to top when new collection is loaded. (Fixes #50)

### LINKED ISSUE #50: Scroll to top when loading new page
Currently the browser stays scrolled down when loading new data into the same kind of page, for example when navigating from one category to another, or from one facet to another.

aslagle:
I'm not sure how to write a test for this.

### DIFF
--- a/src/components/Collection.tsx
+++ b/src/components/Collection.tsx
@@ -87,6 +87,12 @@ export default class Collection extends React.Component<CollectionProps, any> {
     );
   }
 
+  componentWillReceiveProps(nextProps) {
+    if (this.props.isFetching && !nextProps.isFetching) {
+       document.body.scrollTop = 0;
+    }
+  }
+
   componentDidMount() {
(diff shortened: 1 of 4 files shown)


src/components/Collection.tsx
```diff
@@ -87,6 +87,12 @@ export default class Collection extends React.Component<CollectionProps, any> {
     );
   }
 
+  componentWillReceiveProps(nextProps) {
+    if (this.props.isFetching && !nextProps.isFetching) {
```
skomputer: maybe we shouldn't scroll to the top if isFetching is set to false due to an error

skomputer:
we should test this behavior if possible, either by calling componentWillUpdate directly in the test or by updating the props another way (it seems that you can do this with React.render() instead of TestUtils.renderIntoDocument()

skomputer:
:+1:

PR #87, opened 2016-02-29
Publishing and repo organization changes and BookDetailsContainer

aslagle:
This connects to #30.

This has a bunch of changes related to publishing and using the app for the circulation manager admin interface.

I reorganized the repo so there are two npm packages, and a parent package with commands for building. The server depends on the opds-browser, and the build commands will link it for development, but when the server’s deployed by itself it can get the opds-browser from npm and won’t need to build it. I’m not sure they should be in the same repo, but it’s convenient for development. 

I also changed the dev and prod configs of the browser to both build to lib/index.js, which is the main file of the npm package.

I added a config option for a pathFor function to create links, since it’s closely coupled to routing which was already handled outside the main app.

I added a config option to pass in a component to render in place of BookDetails, which will get the BookDetails as a child and can choose what to do with it. I’m using this to make a modal with tabs for details, edit, report a problem, etc. This required making a Modal component separate from BookDetails.

I also changed Link to be compatible with the latest typescript and made the server handle authenticated feeds.


skomputer:
:+1:

PR #91, opened 2016-03-02
Preserve raw entry info in book data.

aslagle:
The BookDetailsContainer may need additional info from the parsed OPDS that the main browser doesn't care about.


aslagle:
This may be useful later, we don't need it for now. We decided to make a separate request for admins to get details about a book instead of adding more info to the main opds entries.

PR #99, opened 2016-03-15
Added function to refresh the current book.

aslagle:
This connects to https://github.com/NYPL-Simplified/circulation/issues/103


skomputer:
:+1:

PR #108, opened 2016-03-30
Better navigation

skomputer:
This branch makes major changes to how the app navigates from one page to another.

Previously, clicking on a link dispatched async actions to fetch new data, update the store, and trigger an onNavigate() callback provided to the app, which updated the web browser's history with a new url. Clicking the back button in a web browser would pass parameters from the browser's history to a function on the app called setCollectionAndBook(), which would fetch new data and update the store. In sum:
```
click link --> fetch new data --> update store --> display new data
           --> trigger callback --> update browser url/history
click back --> setCollectionAndBook() --> fetch new data --> update store --> display new data
```

Now, all page transitions are executed by re-rendering the application with new props. Clicking on a link triggers a navigate() callback provided to the app, which should (but is not required to) update the web browser's history and then re-render the application with props based on the collection and book URLs provided to the callback. Upon receiving new props, the app dispatches async actions to fetch new data, and updates the store. In sum:
```
click link --> trigger callback --> update browser url/history
                                --> render with new props --> fetch new data --> update store --> render data from store
click back --> render with new props --> fetch new data --> update store --> render data from store
```

Hopefully this will make navigation easier to understand and trace.


aslagle:
Looks great, this is much better. The only thing I don't like is isTopLevel getting passed around outside this module- I don't think circulation-web should need to know about it.

aslagle:
:+1:

PR #109, opened 2016-04-01
search doesn't fetch description unless url changes

skomputer:
This branch fixes a bug where the Search component fetches a search description on mount and then receipt of the search description data triggers it to fetch again.


aslagle:
:+1:

PR #112, opened 2016-04-05
Refresh

skomputer:
Recent changes to navigation removed top-level refs and functions but didn't provide a way for BookDetailsContainer to refresh data in the browser, because calling navigate with the existing book and collection url doesn't update data.

This branch resolves this issue by adding a refreshCollectionAndBook() prop to Root, which passes it directly to BookDetailsContainer.


aslagle:
:+1:

PR #113, opened 2016-04-06
Don't try to parse published date if it's undefined.

aslagle:
This fixes a bug where the browser would show "Published: undefined NaN, NaN" for OPDS entries without a published date.


skomputer:
:+1:

PR #114, opened 2016-04-06
DataFetcher uses consistent error types

skomputer:
This branch defines uses new RequestError and RequestRejector interfaces to enforce consistency of error types returned by DataFetcher.


aslagle:
:+1:

PR #115, opened 2016-04-06
Switched to dompurify for html sanitization.

aslagle:
dompurify prevents XSS but allows more tags and attributes than our previous sanitizer.


skomputer:
:+1:

PR #116, opened 2016-04-07
Next prev book

skomputer:
This branch adds keydown event listeners that display the next or previous book in a collection if right or left arrow keys are pressed.


aslagle:
:+1:

PR #124, opened 2016-04-12
interfaces always explicitly imported

skomputer:
This branch moves non-prop interfaces from a declaration file to a new file that exports them and requires them to be explicitly imported elsewhere.


aslagle:
:+1:

PR #130, opened 2016-04-20
retry/refresh bug fix and tests

### LINKED ISSUE #129: retry and refresh fail if collectionUrl or bookUrl are empty

skomputer:
Closes #129.


aslagle:
:+1:

PR #133, opened 2016-04-22
simpler and customizable breadcrumbs

### LINKED ISSUE #134: Descenders get cut off on second line of title in lane
screen shot 2016-04-25 at 1 40 14 pm

skomputer:
This branch simplifies the way history and hierarchy are generated and allows for a custom mapping from history and hierarchy to breadcrumb links.

Also closes #134.


skomputer:
Note to self: breadcrumbs not working as intended when performing a search twice from a subsublane. Note from Winnie to self: breadcrumbs with same title shouldn't be collapsed (I possibly disagree).
One more note to self: fix lane titles for "All Spanish", etc.
Seriously final note: be strong.

aslagle:
If I go to https://circulation.librarysimplified.org, I get "All Books / All Books" as the breadcrumbs. Maybe we need to handle redirects better.

aslagle:
:+1:

PR #138, opened 2016-05-26
Header receives collection and book titles

### LINKED ISSUE #136: Header should receive current feed's title as a prop
It should be possible to display the current feed's title in the header as well as in the breadcrumbs.

skomputer:
Closes #136.


aslagle:
the readme needs to be updated too

aslagle:
:+1:

PR #139, opened 2016-06-07
update dependencies

skomputer:


packages/opds-web-client/src/app.tsx
```diff
@@ -34,8 +34,8 @@ class OPDSWebClient {
       render: function() {
         let { collectionUrl, bookUrl } = this.props.params;
         let mergedProps: RootProps = Object.assign(config, {
-          collectionUrl: collectionUrl ? decodeURIComponent(collectionUrl) : null,
-          bookUrl: bookUrl ? decodeURIComponent(bookUrl) : null
+          collectionUrl,
+          bookUrl
```
aslagle: you could do this because you changed circulation-web to use react-router? would it still work for an app that wasn't using react-router?

aslagle:
:+1:

PR #140, opened 2016-06-08
bind component methods to avoid reundant renders

### LINKED ISSUE #120: bind methods in constructor instead of props
`<Component action={this.action.bind(this)} />` will cause the component to re-render every time, because bind() returns a new function. Instead, this.action should be bound in the parent component's constructor.

skomputer:
Closes #120.


aslagle:
:+1:

PR #141, opened 2016-06-08
lane shows "more" link at end

### LINKED ISSUE #81: add "more" link at end of lane

skomputer:
Closes #81.

### DIFF
--- a/packages/opds-web-client/src/components/BookCover.tsx
+++ b/packages/opds-web-client/src/components/BookCover.tsx
@@ -1,0 +1,44 @@
+import * as React from "react";
+
+export interface BookCoverProps {
+  style?: any;
+  text: string;
+}
+
+export default class BookCover extends React.Component<BookCoverProps, any> {
+  render() {
+    let cellStyle = Object.assign({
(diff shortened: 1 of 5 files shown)


packages/opds-web-client/src/components/BookCover.tsx
```diff
@@ -0,0 +1,44 @@
...
+      textAlign: "center"
+    };
+
+    // text is split so that "More" always appears on its own line
+    return (
+      <div className="bookCover" style={cellStyle}>
+        <div style={contentStyle}>
+          {this.props.text.split(" ")[0]}
+          <br />
+          {this.props.text.split(" ").slice(1).join(" ")}
```
aslagle: how about using props.children for this? so the caller can put "More" on its own line and BookCover doesn't have to know what its content is.

aslagle: maybe not since this component also needs the text to get font size. but if it assumes "More" is the first word, I think it should put the "More" in itself.

skomputer:
Superseded by #142.

PR #145, opened 2016-07-22
Borrow links with auth

skomputer:
This branch adds borrow links to book pages and support for basic authentication.


packages/opds-web-client/src/components/BorrowButton.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+export default class BorrowButton extends React.Component<BorrowButtonProps, any> {
+  constructor(props) {
+    super(props);
+  }
+
+  render() {
+    let props = Object.assign({}, this.props);
+    delete props["book"];
+    delete props["borrow"];
```
aslagle: it doesn't look like this copy of the props is being used

packages/opds-web-client/src/components/DownloadButton.tsx
```diff
@@ -0,0 +1,82 @@
...
+        blob,
+        this.generateFilename(this.props.title),
+        // TODO: use mimeType variable once we fix the link type in our OPDS entries
+        this.mimeType()
+      );
+    });
+  }
+
+  generateFilename(str: string): string {
+    return str.toLowerCase().replace(/[^a-z0-9]+/g, "-").replace(/(^-|-$)/g, "") + this.fileExtension();
```
aslagle: why not use "" in the first replace?

aslagle:
:+1:

PR #146, opened 2016-08-04
added Lanes component

skomputer:


aslagle:
:+1: Having both Lane and Lanes components might be a little confusing but otherwise looks good.

PR #153, opened 2016-08-16
Added a footer with about-type links.

aslagle:
This fixes https://github.com/NYPL-Simplified/circulation-patron-web/issues/22

### DIFF
--- a/packages/opds-web-client/src/OPDSDataAdapter.ts
+++ b/packages/opds-web-client/src/OPDSDataAdapter.ts
@@ -219,6 +219,7 @@ export function feedToCollection(feed: OPDSFeed, feedUrl: string): CollectionDat
   let catalogRootLink: OPDSLink;
   let parentLink: OPDSLink;
   let shelfUrl: string;
+  let aboutLinks: LinkData[] = [];
 
   feed.entries.forEach(entry => {
     if (feed instanceof AcquisitionFeed) {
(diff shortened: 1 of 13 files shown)


packages/opds-web-client/src/OPDSDataAdapter.ts
```diff
@@ -280,6 +281,24 @@ export function feedToCollection(feed: OPDSFeed, feedUrl: string): CollectionDat
     if (shelfLink) {
       shelfUrl = shelfLink.href;
     }
+
+    let aboutLabels = {
+      "about": "About",
+      "terms-of-service": "Terms of Service",
+      "privacy-policy": "Privacy Policy",
+      "copyright": "Copyright"
+    };
```
aslagle: I'm not sure where these should go.

I could've done all this in opds-feed-parser, but it's not really part of OPDS.

packages/opds-web-client/src/components/Root.tsx
```diff
@@ -258,6 +276,17 @@ export class Root extends React.Component<RootProps, any> {
...
+          <footer style={ footerStyle }>
+            <ul aria-label="about links" className="list-inline">
+              { this.props.collectionData.aboutLinks.map(link =>
+                <li key={link.url} style={{padding: "5px"}}>
+                  <a href={link.url}>{link.text}</a>
+                </li>
+              ) }
+            </ul>
+          </footer>
+        }
```
aslagle: I might need to make it possible to pass in a Footer component the way we did with Header. I'm not sure if it makes sense to show these links in the admin interface, and if not, the admin interface could pass in an empty Footer component.

packages/opds-web-client/src/OPDSDataAdapter.ts
aslagle: I looked through some other OPDS feeds and none of them had links like this, so maybe it shouldn't even be here - maybe this should have a configurable footer and circulation-patron-web should handle the links.

courte: circulation-patron-web feels right to me, since these are static-ish pages that exist for the client apps.

courte:
The code appears solid to the untrained eye. I agree that the app-specific links, intended for the end user, should be parsed in patron-web. But, _should_ you decide to leave it here, the code itself appears good to go.

Issue #159, opened 2016-08-24
After borrowing a book, update the loans data in the store with the borrow response


aslagle:
We're currently refreshing the entire loans feed, but it's not necessary and it can be slow.

PR #160, opened 2016-08-24
Update book action

aslagle:
To implement revoke in circulation-patron-web, I need a way to make an authenticated request to the revoke url and process the OPDS entry. I realized the borrow code doesn't actually do anything specific to borrowing, so I changed the action to be more general. You can pass in any url that returns an OPDS entry, optionally after updating a book. This way I can use it for revoke without having to pass any additional props around.

### DIFF
--- a/packages/opds-web-client/src/reducers/loans.ts
+++ b/packages/opds-web-client/src/reducers/loans.ts
@@ -26,6 +26,16 @@ export default (state: LoansState = initialState, action): LoansState => {
         books: action.books
       });
 
+    case "LOAD_UPDATE_BOOK_DATA":
+      // A book has been updated, so the loans feed is now outdated.
+      // If we remove the loans, the components showing the book that
+      // was updated can use the data from the book update request 
+      // until the next LOAD_LOANS action.
+
+      return Object.assign({}, state, {
+        books: []
+      });
+
     default:
       return state;
   }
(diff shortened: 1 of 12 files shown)


packages/opds-web-client/src/reducers/loans.ts
```diff
@@ -26,6 +26,16 @@ export default (state: LoansState = initialState, action): LoansState => {
...
+    case "LOAD_UPDATE_BOOK_DATA":
+      // A book has been updated, so the loans feed is now outdated.
+      // If we remove the loans, the components showing the book that
+      // was updated can use the data from the book update request 
+      // until the next LOAD_LOANS action.
+
+      return Object.assign({}, state, {
+        books: []
+      });
+
```
aslagle: This is the only change that's not just renaming. 

I'm not sure why it wasn't a problem for borrow, but after I used the update book action to revoke a loan, Root would continue using the book data from the loans feed until the loans request completed, but the loading message would go away as soon as the revoke request was done. The loans request can take a while so it would look like your revoke didn't work.

A better way to handle this would be to have this reducer update the loan for that one book, and keep all the other loans. That way we wouldn't need to make another loans request at all. I created an issue for that here https://github.com/NYPL-Simplified/opds-web-client/issues/159.

PR #162, opened 2016-08-31
Replaced jest with mocha and sinon.

aslagle:
This pull request changes the tests to use mocha, chai, and sinon instead of jest, which we're already doing in circulation-patron-web.


packages/opds-web-client/src/DataFetcher.ts
```diff
@@ -90,7 +90,7 @@ export default class DataFetcher {
     options = Object.assign({ credentials: "same-origin" }, options);
 
     if (this.proxyUrl) {
-      let formData = new FormData();
+      let formData = new (window as any).FormData();
```
aslagle: For some reason various globals that worked in jest no longer do in mocha :(

packages/opds-web-client/src/__tests__/state-test.ts
```diff
@@ -1,42 +1,46 @@
...
     }));
-    createFetchCollectionAndBook.mockClear();
+    createFetchCollectionAndBookStub = stub(mergeRootProps, "createFetchCollectionAndBook").returns(fetchCollectionAndBookStub);
+
+    storeStub = stub(store, "default").returns({ dispatch, getState: () => testState });
+  });
+
+  afterEach(() => {
+    createFetchCollectionAndBookStub.restore();
+    storeStub.restore();
```
aslagle: I think this is the main thing I dislike about using sinon + mocha. They aren't integrated so if you stub out a function in an existing module, the stub will persist through all the subsequent tests unless you restore the function.

PR #163, opened 2016-09-01
Moved inline styles to scss.

aslagle:
This branch moves the inline styles for each component into scss files that get pulled in by webpack. I also renamed all the css classes to use dashes instead of camel case


packages/opds-web-client/src/components/BookCover.tsx
```diff
@@ -25,78 +26,42 @@ export default class BookCover extends React.Component<BookCoverProps, any> {
...
-      fontFamily: "Georgia, serif",
-      position: "relative",
-      textAlign: "left"
-    };
 
     return (
-      <div className="autoBookCover" style={cellStyle}>
-        <div style={titleStyle}>{ title }</div>
-        <div style={authorStyle}>{ authors.join(", ") }</div>
+      <div className="auto-book-cover" style={{ backgroundColor: bgColor }}>
```
mgiraldo: any reason why you decided to keep the style here? (and the next few lines)

packages/opds-web-client/src/components/LaneMoreLink.tsx
```diff
@@ -1,49 +1,27 @@
...
+    let fontSize = this.computeFontSize();
 
     return (
       <CatalogLink
-        className="moreLink"
-        style={moreLinkStyle}
+        className="more-link"
         collectionUrl={this.props.lane.url}>
-        <div style={cellStyle}>
+        <div style={{ fontSize: fontSize }}>
```
mgiraldo: another rogue style?

packages/opds-web-client/src/components/BookCover.tsx
aslagle: These are computed based on the title and author, so I don't think it's possible to do it outside the js.

packages/opds-web-client/src/components/LaneMoreLink.tsx
aslagle: This is another one that's computed based on the text in it..

packages/opds-web-client/src/components/BookCover.tsx
mgiraldo: oh i see it's like the cover generator

aslagle: yeah, it picks a random color for the cover and tries to pick a font size that will make the title and author fit.

packages/opds-web-client/src/components/LaneMoreLink.tsx
aslagle: It's the link at the end of a lane when you scroll all the way to the right that's the size of a book cover and says "More Young Adult Fiction" or whatever the lane name is.

PR #169, opened 2016-10-03
Multiple auth providers

aslagle:
This pr will enable logging in with clever in circulation-patron-web.

OPDSWebClient can receive a list of authPlugins that implement the AuthPlugin interface, and it will show buttons for the user to select between them. Each plugin can implement a button for that screen, and a custom form if the button does not link to another site. This package includes a single auth plugin for basic auth. 

I also added a cancel button to the auth forms, and modified the auth middleware to support this. The auth middleware didn't have any tests so I added that too.

### DIFF
--- a/packages/opds-web-client/src/components/Root.tsx
+++ b/packages/opds-web-client/src/components/Root.tsx
@@ -8,7 +8,7 @@ import {
 import BookDetails from "./BookDetails";
 import LoadingIndicator from "./LoadingIndicator";
 import ErrorMessage from "./ErrorMessage";
-import BasicAuthForm  from "./BasicAuthForm";
+import AuthProviderSelectionForm  from "./AuthProviderSelectionForm";
 import Search from "./Search";
 import Breadcrumbs, {
   ComputeBreadcrumbs, defaultComputeBreadcrumbs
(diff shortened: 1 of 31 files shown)


packages/opds-web-client/src/components/Root.tsx
```diff
@@ -67,10 +67,10 @@ export interface RootProps extends StateProps {
   fulfillBook?: (url: string) => Promise<Blob>;
   indirectFulfillBook?: (url: string, type: string) => Promise<string>;
   fetchLoans?: (url: string) => Promise<any>;
-  saveBasicAuthCredentials?: (credentials: string) => void;
-  clearBasicAuthCredentials?: () => void;
-  showBasicAuthForm?: (callback: BasicAuthCallback, labels: BasicAuthLabels, title: string) => void;
-  closeErrorAndHideBasicAuthForm?: () => void;
+  saveAuthCredentials?: (credentials: AuthCredentials) => void;
+  clearAuthCredentials?: () => void;
+  showAuthForm?: (callback: AuthCallback, providers: any, title: string) => void;
```
winniequinn: Can we make `any` a more specific type?

packages/opds-web-client/src/interfaces.ts
```diff
@@ -134,18 +136,29 @@ export interface NavigateContext {
...
+
+export interface AuthCallback {
+  (): any;
 }
 
-export interface BasicAuthData {
+export interface AuthProvider {
+  name: string;
+  plugin: AuthPlugin;
+  method: any;
```
winniequinn: What should the type of `method` be here?

PR #180, opened 2017-01-17
Accessibility improvements

aslagle:
@winniequinn

### DIFF
--- a/packages/opds-web-client/src/components/Collection.tsx
+++ b/packages/opds-web-client/src/components/Collection.tsx
@@ -53,8 +53,8 @@ export default class Collection extends React.Component<CollectionProps, any> {
 
           { this.props.collection.books &&
             <ul aria-label="books" className="subtle-list books">
-            { this.props.collection.books.map(book =>
-              <li key={book.id}>
+            { this.props.collection.books.map((book, index) =>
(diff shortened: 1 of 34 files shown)


packages/opds-web-client/src/components/Collection.tsx
```diff
@@ -102,24 +101,25 @@ export default class Collection extends React.Component<CollectionProps, any> {
 
   componentWillReceiveProps(nextProps) {
     if (this.props.isFetching && !nextProps.isFetching && !nextProps.error) {
-       document.body.scrollTop = 0;
-       document.documentElement.scrollTop = 0;
+      (this.refs["body"] as any).scrollTop = 0;
```
winniequinn: @aslagle Out of curiosity, why was this change needed? Does `document.body.scrollTop` not work as intended?

winniequinn: @aslagle Can we maybe use `as HTMLElement` for this cast and similar casts below?

aslagle: This was needed because document.body is no longer the element that scrolls. I renamed the body ref so this is more clear. I was able to change some of the casts to HTMLElement.

PR #195, opened 2017-05-24
Added an option to specify an epub reader URL template, and incorporated EDRLab's webpub streamer and NYPL's webpub navigator into the demo server.

aslagle:
On the demo server, any books that have an open access EPUB URL will have a "Read" button that opens up the reader with a manifest from the streamer.


danielweck:
great :)

PR #199, opened 2017-09-14
List view

### LINKED ISSUE #196: Add "List" view option to the current  "Grid" view of books
In the current OPDS Web Catalog, there is no traditional listing of book items other than a collection of grouped feeds organized as swim lanes, grid of book covers with facets and filters and a single item book detail page. 

The mobile app catalogs have swim lanes, and list views and book detail views.

A conceptual wire frame layout is attached to be more reflective of the type of catalog views used for similar web catalogs such as NYU, NYPL and commercial vendors like Overdrive and Bibliotheca.

https://docs.google.com/presentation/d/1XTSDNA5TfeXNHcF8Hh7iPHZxhVQ9TpTmq3VemcRWz_M/edit#slide=id.p

aslagle:
This fixes #196: First pass implementation of the list view. 

I also changed a little of the language - "Open in Web Reader" => "Read Online", and "Get" => "Borrow" because there has to be more than one step to get the book, so it shouldn't match the app.

### DIFF
--- a/packages/opds-web-client/src/components/Book.tsx
+++ b/packages/opds-web-client/src/components/Book.tsx
@@ -217,7 +217,8 @@ export default class Book<P extends BookProps> extends React.Component<P, void> 
         <button key="onhold" className="btn btn-default disabled">Reserved</button>
       );
     } else if (!this.isBorrowed() && this.props.book.borrowUrl) {
-      let label = this.props.book.copies &&
+      let label = !this.isReady() &&
+                  this.props.book.copies &&
                   this.props.book.copies.available === 0 ?
                   "Reserve" :
                   "Borrow";
@@ -240,6 +241,11 @@ export default class Book<P extends BookProps> extends React.Component<P, void> 
   isReserved() {
     return this.props.book.availability &&
            this.props.book.availability.status === "reserved";
+  }
+
+  isReady() {
+    return this.props.book.availability &&
+           this.props.book.availability.status === "ready";
   }
 
   isBorrowed() {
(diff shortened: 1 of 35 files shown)


packages/opds-web-client/src/components/Book.tsx
```diff
@@ -1,15 +1,29 @@
...
 
-export default class Book extends React.Component<BookProps, void> {
+export default class Book<P extends BookProps> extends React.Component<P, void> {
+  constructor(props) {
+    super(props);
+    this.borrow = this.borrow.bind(this);
+  }
+
   render(): JSX.Element {
+    let summary = (this.props.book && this.props.book.summary && this.props.book.summary.replace(/<\/?[^>]+(>|$)/g, " ")) || "";
```
leonardr: This is checking that the summary is more than just HTML tags, right? Not actually removing HTML tags from the summary.

aslagle: No, it's actually removing the HTML tags, in order to provide a more compact view of the summary without any line breaks or styling that might be in the HTML.

leonardr: Then is `dangerouslySetInnerHTML={{__html: summary}}` just to be extra safe?

I don't have a strong opinion against stripping HTML out of the summary, it's just not what I was expecting.

aslagle: You're right, dangerouslySetInnerHTML is probably not necessary, I can remove that.

I'm stripping it because I noticed a bunch of books that started with 3 or 4 lines of like An Amazon Best Book of 2014, A Library Journal Best Book of 2014, etc and in list view that would end up as the entire summary. I wanted to fit as much as possible of the actual description in. You can click "More" to see the nicely formatted summary.

aslagle: Actually dangerouslySetInnerHTML is still needed to render HTML character entities which I'm not stripping out or changing.

PR #206, opened 2017-11-24
Cleanup

aslagle:
Updated a few dependencies and fixed warnings.


packages/opds-web-client/src/app.tsx
```diff
@@ -23,12 +25,12 @@ class OPDSWebClient {
   }, elementId: string) {
     this.elementId = elementId;
     this.pathPattern = config.pathPattern || "/(collection/:collectionUrl/)(book/:bookUrl/)";
-    this.RouteHandler = React.createClass({
+    this.RouteHandler = createReactClass({
```
EdwinGuzman: Any plans on moving this soon to the `React.component` pattern?

aslagle: We can if you want :)

I think this is the only place it's used.

EdwinGuzman: Might be good to also remove the `react-create-class` dependency.

aslagle: Agreed.

PR #207, opened 2017-11-24
More cleanup

aslagle:
Some additional cleanup and class-level comments for many of the components.

### DIFF
--- a/packages/opds-web-client/src/AuthPlugin.ts
+++ b/packages/opds-web-client/src/AuthPlugin.ts
@@ -1,6 +1,9 @@
 import { AuthFormProps, AuthButtonProps } from "./components/AuthProviderSelectionForm";
 import { AuthMethod, AuthCredentials } from "./interfaces";
 
+/** Applications can implement this interface if they would like to support authentication
+    methods other than basic auth. A list of `AuthPlugin`s can be passed as a prop to the
+    `OPDSCatalog` component. */
 interface AuthPlugin {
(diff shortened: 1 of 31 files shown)


aslagle:
@EdwinGuzman let me know if these comments are helpful or if there's something better I could be doing.

EdwinGuzman:
Looks good on a first glance. Are you following a specific documentation syntax, like esdoc?

aslagle:
I was trying to be compatible with http://typedoc.org/ but the comments don't have much detail.

EdwinGuzman:
Thanks, I'll looking into that.

packages/opds-web-client/src/AuthPlugin.ts
```diff
@@ -1,6 +1,9 @@
 import { AuthFormProps, AuthButtonProps } from "./components/AuthProviderSelectionForm";
 import { AuthMethod, AuthCredentials } from "./interfaces";
 
+/** Applications can implement this interface if they would like to support authentication
+    methods other than basic auth. A list of `AuthPlugin`s can be passed as a prop to the
```
EdwinGuzman: Don't you need to start new lines with `*` and start the comment after the first line? Unless if this gets read the same way?

aslagle: I haven't actually tried running this through typedoc, but it's a valid comment. I'm not sure.

Issue #185, opened 2017-02-13
Remove link from breadcrumb for current page and change breadcrumb separator


aslagle:
See https://nypl.github.io/design-toolkit/layout.html#pageHeader

PR #212, opened 2017-11-28
Forcing the breadcrumb li to render a greater than sign instead of a slash.

### LINKED ISSUE #211: Use ">" for breadcrumbs instead of "/"

EdwinGuzman:
Fixes #211 by overriding what Bootstrap automatically adds as a breadcrumb separator.


EdwinGuzman:
Updated the PR to also fix #185 which is closely related.

aslagle:
@EdwinGuzman this looks great. There is one place where the last breadcrumb maybe should be a link though - when you're on the book details page, it's useful to be able to go back to the page you came from.

EdwinGuzman:
Yeah, I just noticed that as well. If we follow this pattern, then it'd be good to have a "back" button on the book details page instead of updating the breadcrumb for this use case. Otherwise, I'd imagine that it'll be tricky to let the breadcrumb know that it's on the book details page?

aslagle:
I think you could pass it a prop from Root to say whether to make the last one a link. Root renders the book so it knows when you're on that page.

EdwinGuzman:
Updated and got tests working, too.

packages/opds-web-client/src/components/__tests__/Breadcrumbs-test.tsx
```diff
@@ -30,12 +30,29 @@ describe("Breadcrumbs", () => {
...
+    // The current page is no longer a link unless `currentLink` is true:
+    expect(links.length).to.equal(1);
     expect(links.at(0).props().children).to.contain("2nd title");
     expect(links.at(0).props().collectionUrl).to.equal("2nd url");
     // last link is wrapped in <strong>
-    expect((links.at(1).props().children as any).props.children).to.contain("last title");
+    expect(currentLink.text()).to.equal("last title");
+  });
+
+  it("shows links with bootstrap classes", () => {
```
aslagle: I would rename these two tests to reflect that they're testing with and without the current link now.

aslagle:
Works great, this is good to merge if you update those test names. Could you also change the version number? And tomorrow I'll add you to the npm package so you can publish :)

PR #216, opened 2018-04-20
Media icon item cover

### LINKED ISSUE #215: Indicate media type in list view
When browsing a mixed list of ebooks and audiobooks, some graphical (but accessible) indicator should be used to make the media type (ebook, audiobook, etc) obvious without having to click through to a detail page.

EdwinGuzman:
Fixes #215 

Adding `@nypl/dgx-svg-icons` package and updating the styles for the item cover to include either a headphone or book icon. Also doing the same in the item details page.

Screenshots for @ricardoom because I'm not sure if he can run it locally (but can help next week with that):

Item details
screen shot 2018-04-20 at 5 03 08 pm

Lanes
screen shot 2018-04-20 at 4 44 00 pm


aslagle:
@EdwinGuzman this repo is meant to work with _any_ OPDS feed, not just the circulation manager. And the media type is not part of the OPDS spec, so other OPDS feeds won't have it. I'm okay with having it in here, but things need to continue to work well when there are no media types in the OPDS feed. I'd also prefer to parse the media type as part of opds-feed-parser if we're going to use it here.

EdwinGuzman:
@aslagle that makes a lot of sense. Thanks.
I updated the style issue for the svg and added it to the list view. I also checked to see if it is in the opds feed and if not it won't render or mess up the styles without it on the cover. So now it should work for both example NYPL's circ managaer and Standard Ebooks feeds.

Updating the opds-feed-parser sounds like the better solution and I'll take a look at that too. That'll also help with getting the media in `circulation-web`, right?

aslagle:
Yeah, if you parse it in `opds-feed-parser` you'll be able to use that in `circulation-web` as well.

ricardoom:
@aslagle I want to leave the labels for the media type as is, its bad practice to rely on text, image, or color only to provide meaning, really it should be a combination of at least 2. 

I'm open to adding _more_ info there, but what would that be? I thought this issue was to allow for users to quickly scan the lane and see what format an item was without having to hover or click through. 

Presently the hover state has no transparency (correct me if i'm wrong) and covers the entire book. I presume the idea here is to show them the title and author name. I think if anything the transparency should be removed and have "toast" pop up  overlay be fully opaque. Doing so will maximize readability.

Now That I'm looking at them as a lane it might be better to apply the transparency to the media type bar, and make the full toast completely opaque. So reversing the transparency for the label and the pop up might be a better way to proceed.

aslagle:
@ricardoom I was thinking it would be nice to have something to indicate whether the book is currently available or how long the holds queue is. If we're going to keep the text I'd like to change the capitalization of "EBook". I'm not really sure what it should be but that looks wrong.

In the old version the hover state is fully opaque and doesn't cover the whole book - the top of the cover still shows above the popup thing. It also has the title and author in the theme color rather than gray, which I liked.

I would agree that it's better to have the whole thing be opaque, and I'm not sure the media bar should go to the top - maybe the other part could go up but leave the bar at the bottom.

aslagle:
This looks good. I'd vote for "eBook", "Audio", and "PDF" but I'm not sure what we use elsewhere.

EdwinGuzman:
Those are good. Just want to point out that for screenreaders, I think it should be "e book", "audio", and "p d f" or else they'll be read incorrectly.

aslagle:
Now that the hover state is opaque I think it needs something to distinguish it from the page background - either a border or a slightly different color.

packages/opds-web-client/src/components/Book.tsx
```diff
@@ -83,6 +90,11 @@ export default class Book<P extends BookProps> extends React.Component<P, void>
               { this.fields().map(field =>
                 field.value ? <div className={field.name.toLowerCase().replace(" ", "-")} key={field.name}>{field.name}: {field.value}</div> : null
               ) }
+              {
+                this.getMedium(this.props.book) && (
+                  <span>Media: {this.getMediumSVG(this.getMedium(this.props.book))}</span>
```
aslagle: I'm not sure it needs the "Media: " prefix, I think the SVG and text is enough here.

packages/opds-web-client/src/components/Book.tsx
```diff
@@ -237,6 +249,38 @@ export default class Book<P extends BookProps> extends React.Component<P, void>
     return links;
   }
 
+  getMedium(book) {
+    if (!book.raw || !book.raw["$"]) {
+      return "";
+    }
+
+    return book.raw["$"]["schema:additionalType"].value;
```
aslagle: It's also possible `book.raw["$"]` exists but does not contain `"scheme:additionalType"`.

packages/opds-web-client/src/components/__tests__/Book-test.tsx
```diff
@@ -108,6 +110,23 @@ describe("Book", () => {
...
+          fulfillBook={fulfillBook}
+          indirectFulfillBook={indirectFulfillBook}
+        />
+      );
+
+      let itemIcon = wrapper.find(".item-icon");
+      let svg = itemIcon.find(AudioHeadphoneIcon);
+
+      expect(svg.length).to.equal(2);
+      expect(itemIcon.first().render().text()).to.equal("NYPL Audio/Headphone SVG Icon Audio");
```
aslagle: I didn't realize the dgx icons said "NYPL" in them. That could be misleading for an OPDS feed that has nothing to do with NYPL.

packages/opds-web-client/src/stylesheets/book.scss
```diff
@@ -43,8 +51,9 @@
   margin-bottom: $padding;
 
   .compact-info {
-    background-color: rgba($pagecolor, 0.8);
-    border-top: 2px solid $linkcolor;
+    background-color: rgba($pagecolor, 0.75);
+    border-top: 2px solid #000;
+    color: #000;
```
aslagle: I'd like to change these `#000`s back to `$linkcolor`.

packages/opds-web-client/src/components/Book.tsx
EdwinGuzman: I thought the same but I added it to be consistent with the previous information layout.

EdwinGuzman:
Okay, made all the updates. I'll leave it up to @ricardoom to decide on the theme color for the text. Also, not sure if that color combination is accessible on the slightly transparent background.

ricardoom:
I'd prefer to move away entirely from the default Bootstrap Blue, to a very simple monochrome... but i know that could be a bit of a headache, @aslagle what say you?

aslagle:
I tried to set it up so that other apps/libraries could pick their own theme colors when building on this repo. I'd be fine with switching to different colors for the neutral theme in here, but the colors need to be in _variables.scss so that other apps can easily override them. I don't think that would be a headache, should be pretty straightforward.

EdwinGuzman:
You can create a list with both audio and ebooks so the case of a lane with ebooks and audiobooks should come up.

Do you mean icons in the entry point options when creating a list? I can add that but that'd be in `circulation-web`.

It may be too much to have every book say `eBook`, but we should not have standalone icons convey meaning without text.

aslagle:
No, I mean in the new design for having entry points at the top of the main catalog and search results. I think if the entry points were at the top to connect the icons to the text it wouldn't be necessary to repeat the text every place the icon is used.

EdwinGuzman:
Hmm I'm not sure I've seen that new design yet...

aslagle:
I'm not sure there was one for the web app, it was for the mobile apps. But presumably we'd do something similar.

EdwinGuzman:
Got it. Then we'd have to wait for designs from @ricardoom .

EdwinGuzman:
I was thinking about this, the entry points will not necessarily be media types since we can later have a `Research` entry point. So there may not be an icon for each entry point. I briefly talked to @ricardoom about this and I'm still a bit unclear about any existing designs for the web app.

aslagle:
Yeah I guess that's true, and some libraries may not turn on the entry points at all. But I would like to come up with a way to have the icons + text in a prominent place and not have to repeat the text so much.

ricardoom:
https://sketch.cloud/s/nmjlg/all/page-1/opds-web-desktop
here is an updated version
screen shot 2018-05-10 at 5 24 05 pm

Still rough around the edges. But the basics are there and I think can be worked on. @aslagle feedback please so I can keep this rolling...

aslagle:
@ricardoom This looks good, I'd also want to see what the hover state looks like though.

I also think there are a few different cases we should consider for the media types:
- Everything in the current feed has the same media type and maybe there are links to feeds with other media types. I like what you have for this.
- The current feed has a mixture of eBooks and audio or other types, and may or may not have links, so we might not have the tabs.
- We don't know the media types at all. For that we probably just leave out the icons.

EdwinGuzman:
I believe the hover state would stay the same as it is in this PR but now the overlay won't read what media type an item is.

My thinking was that the tabs are entry points and not media types. So the `Research` entry point could potentially have `ebook` and `audio` media type mix.

EdwinGuzman:
@aslagle @ricardoom I believe I am done with the opds-web-client part of this update.

I will be making another PR soon for circulation-web for the entry points tab:

screen shot 2018-06-20 at 2 48 02 pm

packages/opds-web-client/src/components/Book.tsx
```diff
@@ -237,6 +252,41 @@ export default class Book<P extends BookProps> extends React.Component<P, void>
...
+  }
+
+  getMediumSVG(medium, displayLabel = true) {
+    if (!medium) {
+      return null;
+    }
+
+    const svgMediumTypes = {
+      "http://bib.schema.org/Audiobook": {
+        element: <AudioHeadphoneIcon ariaHidden title="Audio/Headphone SVG Icon" />,
```
aslagle: I'm curious why "SVG" is part of the title?

packages/opds-web-client/src/components/Root.tsx
```diff
@@ -64,6 +71,7 @@ export interface RootProps extends StateProps {
   fetchPage?: (url: string) => Promise<CollectionData>;
   Header?: new() => __React.Component<HeaderProps, any>;
   Footer?: new() => __React.Component<FooterProps, any>;
+  EntryPointsTabs?: new() => __React.Component<EntryPointsTabsProps, any>;
```
aslagle: I would rather have this prop be something more generic, like "CollectionHeader". You could put anything in there, it just happens to be entry points for the circ manager.

packages/opds-web-client/src/components/Root.tsx
```diff
@@ -113,6 +122,7 @@ export class Root extends React.Component<RootProps, RootState> {
     let showBreadcrumbs = this.props.collectionData && breadcrumbsLinks.length > 0;
     let showSearch = this.props.collectionData && this.props.collectionData.search;
     let showFooter = this.props.collectionData && Footer;
+    let showEntryPointsTabs = !!EntryPointsTabs;
```
aslagle: I think the entry points should only be displayed if you're showing a collection rather than a book.

aslagle: I guess you're doing something with the bookPage prop, maybe I'll have to wait for the next pr :)

packages/opds-web-client/src/components/Root.tsx
```diff
@@ -37,6 +37,13 @@ export interface FooterProps extends React.Props<any> {
   collection: CollectionData;
 }
 
+export interface EntryPointsTabsProps extends React.Props<any> {
+  list?: any[];
```
aslagle: Looks like you didn't use this prop.

packages/opds-web-client/src/components/__tests__/Root-test.tsx
```diff
@@ -755,4 +756,71 @@ describe("Root", () => {
       expect(push.args[0][0]).to.equal(context.pathFor(collectionUrl, null));
     });
   });
-});
\ No newline at end of file
+
+  describe("provided a EntryPointsTabs", () => {
+    class Tabs extends React.Component<EntryPointsTabsProps, any> {
```
aslagle: This is a nitpick but that `any` can be `void` here.

packages/opds-web-client/src/stylesheets/book.scss
```diff
@@ -1,12 +1,14 @@
-@import "variables";
```
aslagle: How does it work without this?

packages/opds-web-client/src/components/__tests__/Root-test.tsx
EdwinGuzman: I'm still learning Typescript, so don't worry about being nitpicky.

packages/opds-web-client/src/stylesheets/book.scss
EdwinGuzman: I moved `variables` and `mixins` to the top of `app.scss` so that all the files can have access to it. I _think_ that by importing `variables` into multiple files, it'll be added and compiled multiple times.

packages/opds-web-client/src/components/__tests__/Root-test.tsx
aslagle: That second type is for the state, and this component has no state.

packages/opds-web-client/src/components/Root.tsx
EdwinGuzman: Yes, but now that I think about it, it makes more sense to just not render the component as a whole instead of rendering it but then checking if a book is being displayed to not render parts of it.

packages/opds-web-client/src/stylesheets/book.scss
aslagle: I think that might break circulation-patron-web as it's currently written. It defines a bunch of its own variables and then imports "app.scss", see https://github.com/NYPL-Simplified/circulation-patron-web/blob/master/src/themes/nypl.scss.

packages/opds-web-client/src/components/Book.tsx
EdwinGuzman: No real reason and it can be removed. It was just the convention we used here but we don't need to point out that it's SVG to screenreaders, especially if we're hiding it.

packages/opds-web-client/src/stylesheets/book.scss
EdwinGuzman: I'll test it there to make sure.

EdwinGuzman: I tested it and it works...

But I ran into this error: `../opds-web-client/packages/opds-web-client/lib/components/Breadcrumbs.d.ts(12,26): error TS2304: Cannot find name 'Partial'.`

But I updated the TypeScript version in `circulation-patron-web` and it was fine. So I'll send a PR for that too, since it could also use a version bump for `opds-web-client`.

EdwinGuzman:
Updated feedback comments.

EdwinGuzman:
Updated.

EdwinGuzman:
@aslagle one more update, now I'm just passing the `facetGroups` data to the `CollectionHeader` component and getting all the needed data from there.

packages/opds-web-client/src/components/Book.tsx
```diff
@@ -83,6 +93,11 @@ export default class Book<P extends BookProps> extends React.Component<P, void>
               { this.fields().map(field =>
                 field.value ? <div className={field.name.toLowerCase().replace(" ", "-")} key={field.name}>{field.name}: {field.value}</div> : null
               ) }
+              {
+                bookMedium && (
+                  <span>{this.getMediumSVG(bookMedium, false)}</span>
```
aslagle: I think this might look better with the label? And I'd probably put it above the publisher.

packages/opds-web-client/src/components/Root.tsx
```diff
@@ -113,6 +119,12 @@ export class Root extends React.Component<RootProps, RootState> {
     let showBreadcrumbs = this.props.collectionData && breadcrumbsLinks.length > 0;
     let showSearch = this.props.collectionData && this.props.collectionData.search;
     let showFooter = this.props.collectionData && Footer;
+    // The tabs should only display if the component is passed and if
+    // the catalog is being displayed and not a book.
+    let showCollectionHeader = !!CollectionHeader && !showBook;
+    let facetGroups = this.props.collectionData ?
+      this.props.collectionData.facetGroups : [];
+    let rootUrl = this.props.collectionData ? this.props.collectionData.url : "";
```
aslagle: Looks like you didn't end up needing this variable.

packages/opds-web-client/src/components/Book.tsx
EdwinGuzman: Updated.

packages/opds-web-client/src/components/Root.tsx
EdwinGuzman: Removed it.

packages/opds-web-client/src/stylesheets/book.scss
```diff
@@ -43,28 +59,67 @@
...
 
   .compact-info {
-    background-color: rgba($pagecolor, 0.8);
-    border-top: 2px solid $linkcolor;
+    display: flex;
+    flex-direction: column;
+    background-color: rgba($pagecolor, 0.75);
+    border: 1px solid $pagetextcolor;
+    border-top: 2px solid;
+    color: $pagetextcolor;
```
aslagle: I think I would like this better if it was the link color instead of the page text color - and it is a link.

packages/opds-web-client/src/components/Book.tsx
```diff
@@ -27,27 +31,33 @@ export default class Book<P extends BookProps> extends React.Component<P, void>
...
-            }
-            <div className="authors">
-              {
-                this.props.book.authors.length ?
-                this.props.book.authors.join(", ") :
-                  this.props.book.contributors && this.props.book.contributors.length ?
-                  this.props.book.contributors.join(", ") :
-                  ""
+          <div className={`compact-info ${showMediaIconClass}`}>
+            {this.getMediumSVG(bookMedium, false)}
```
aslagle: I showed this to Mauricio and he suggested putting the title next to this svg in compact view, and hiding it when you hover.

something like:
```
<div class="item-icon"><svg>{the icon here}</svg><span class="brief">Tibetan Peach Pie: A True Account of an Imaginative Life    </span></div>

.item-icon {
    display: flex;
}
.brief {
    white-space: nowrap;
    overflow: hidden;
    text-overflow: ellipsis;
    width: calc(100% - 30px);
    line-height: 1.8;
}
```


image

packages/opds-web-client/src/stylesheets/book.scss
EdwinGuzman: I see, but since the dark color was in @ricardoom 's mockup, I'm leaving it as is.

EdwinGuzman: It is a link, maybe it could be underlined to better represent that. Thoughts @ricardoom ?

packages/opds-web-client/src/components/Book.tsx
EdwinGuzman: I like this. It's not part of @ricardoom 's mockup or the criteria for this ticket so I don't think it should be included, at least not for this iteration, or else we won't be moving along with anything.

ricardoom: I'd say I'm not crazy about putting the title in there, I just dont see how its helpful in the collapsed state. Inevitably the title will be truncated unless the card size is doubled, but I dont think that is realistic. I'd say revert back to icon only, and let the user do some hovering and discovering...

aslagle: I don't know that the title is particularly helpful, but I don't like how empty it looks with only the icon.

Also @EdwinGuzman, sorry but I didn't have an opportunity to weigh in on the mockup or the criteria for this ticket until this PR, so I think it's reasonable to have this discussion now.

EdwinGuzman:
@aslagle anything else you think this PR needs?

aslagle:
I think this mostly looks good now but the way the focus outline is mostly cut off by the border is a little weird:
screen shot 2018-07-03 at 9 37 22 pm

Maybe the border should use the focus color?

EdwinGuzman:
The border should use the focus color, but do you mean that all the time or only when it's focused?

aslagle:
Only when it’s focused.

EdwinGuzman:
Gotcha. Also make the borders in the middle orange since it looked off when it was black and focus outline was orange.

aslagle:
While testing this I noticed an unrelated bug where the "More..." link in the summary in list view can show up when you tab to it even if it is supposed to be hidden.

EdwinGuzman:
Do you mean this "more" link? It's always visible for me. Or do you mean on mobile?

screen shot 2018-07-05 at 12 10 24 pm

EdwinGuzman:
I see. I made a note of it and will create a ticket with that and other possible UI changes.

PR #219, opened 2018-07-23
Wrapping the Collection component instead of rendering the added Coll…

EdwinGuzman:
…ection wrapper separately.

Updated the name of the prop and component and wrapped it around the `Collection` component instead of rendering it separately.

This is part of a refactoring for [circulation-web #159](https://github.com/NYPL-Simplified/circulation-web/pull/159).


EdwinGuzman:
Thanks.

PR #220, opened 2018-09-09
Fixed JS error when there are no books in a lane.

aslagle:


aslagle:
One that I know of is the related books lanes in circulation-patron-web. For example when a book has no books by the same author, the author feed will have that book,  but the client will filter it out and the lane will have no visible books.

PR #221, opened 2018-09-20
Adding 'By' before an author(s) name.

EdwinGuzman:
I thought I completed this in a previous PR but I missed it. This covers JIRA ticket [SIMPLY-167](https://jira.nypl.org/browse/SIMPLY-617) - adds "by" before an author(s) name.

@aslagle , found three different cases:

1)
screen shot 2018-09-20 at 3 47 06 pm
2) 
screen shot 2018-09-20 at 3 47 11 pm
3)
screen shot 2018-09-20 at 3 47 26 pm

### DIFF
--- a/packages/opds-web-client/package.json
+++ b/packages/opds-web-client/package.json
@@ -1,6 +1,6 @@
 {
   "name": "opds-web-client",
-  "version": "0.1.21",
+  "version": "0.1.22",
   "description": "OPDS web client",
   "main": "lib/index.js",
   "typings": "lib/index.d.ts",
--- a/packages/opds-web-client/package.json
+++ b/packages/opds-web-client/package.json
@@ -1,6 +1,6 @@
 {
   "name": "opds-web-client",
-  "version": "0.1.20",
+  "version": "0.1.21",
   "description": "OPDS web client",
   "main": "lib/index.js",
   "typings": "lib/index.d.ts",
--- a/packages/opds-web-client/src/components/Book.tsx
+++ b/packages/opds-web-client/src/components/Book.tsx
@@ -28,38 +28,41 @@ export default class Book<P extends BookProps> extends React.Component<P, void> 
   }
 
   render(): JSX.Element {
+    const book = this.props.book;
     // Remove HTML tags from the summary to fit more information into a truncated view.
     // The summary may still contain HTML character entities and needs to be rendered as HTML.
-    let summary = (this.props.book && this.props.book.summary && this.props.book.summary.replace(/<\/?[^>]+(>|$)/g, " ")) || "";
-    const bookMedium = this.getMedium(this.props.book);
+    let summary = (book && book.summary && book.summary.replace(/<\/?[^>]+(>|$)/g, " ")) || "";
+    const bookMedium = this.getMedium(book);
     const showMediaIconClass = bookMedium ? "show-media" : "";
+    const hasAuthors = !!(book.authors && book.authors.length);
+    const hasContributors = !!(book.contributors && book.contributors.length);
+    const contributors = book.contributors && book.contributors.length ?
+      book.contributors.join(", ") :
+      "";
+    // Display contributors only if there are no authors.
+    const authors = hasAuthors ? book.authors.join(", ") : contributors;
 
     return (
-      <div className={`book ${showMediaIconClass}`} lang={this.props.book.language}>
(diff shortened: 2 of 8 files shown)


packages/opds-web-client/src/components/Book.tsx
```diff
@@ -51,6 +51,7 @@ export default class Book<P extends BookProps> extends React.Component<P, void>
                 <div className="series">{this.props.book.series.name}</div>
               }
               <div className="authors">
+                <span>By </span>
```
aslagle: This probably shouldn’t show if there are no authors and no contributors.

EdwinGuzman: Updated.

packages/opds-web-client/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "opds-web-client",
-  "version": "0.1.20",
+  "version": "0.1.21",
```
aslagle: I think I might have used this version number so you might need to update it again.

EdwinGuzman: Forgot to pull, thanks.

packages/opds-web-client/src/components/Book.tsx
```diff
@@ -28,38 +28,41 @@ export default class Book<P extends BookProps> extends React.Component<P, void>
   }
 
   render(): JSX.Element {
+    const book = this.props.book;
     // Remove HTML tags from the summary to fit more information into a truncated view.
     // The summary may still contain HTML character entities and needs to be rendered as HTML.
-    let summary = (this.props.book && this.props.book.summary && this.props.book.summary.replace(/<\/?[^>]+(>|$)/g, " ")) || "";
-    const bookMedium = this.getMedium(this.props.book);
+    let summary = (book && book.summary && book.summary.replace(/<\/?[^>]+(>|$)/g, " ")) || "";
```
adriana-alter: What's the purpose of this line?

EdwinGuzman: Looks like it cleans up the summary.

packages/opds-web-client/src/components/BookCover.tsx
```diff
@@ -53,7 +53,9 @@ export default class BookCover extends React.Component<BookCoverProps, BookCover
     return (
       <div className="auto-book-cover" style={{ backgroundColor: bgColor }}>
         <div className="title" style={{ fontSize: titleFontSize }}>{ title }</div>
-        <div className="authors" style={{ fontSize: authorFontSize }}>By { authors.join(", ") }</div>
+        { authors.length &&
+          <div className="authors" style={{ fontSize: authorFontSize }}>By { authors.join(", ") }</div>
+        }
```
adriana-alter: Just out of curiosity, what's the authorFontSize thing?  Is the font size in which authors' names are displayed determined by how many authors there are...or...?

packages/opds-web-client/src/components/Book.tsx
adriana-alter: What kind of thing would there be that would need cleaning up?

EdwinGuzman: @aslagle ?

packages/opds-web-client/src/components/BookCover.tsx
EdwinGuzman: Yes, that seems to be correct. The bigger the word count the smaller the font size.

packages/opds-web-client/src/components/Book.tsx
aslagle: This is in the compact list view where we're trying to fit as much info as possible. Some of the distributors put lots of line/paragraph breaks in their summaries which means hardly any text ends up in the compact view- sometimes it's just one line and then empty space. This strips out all of that formatting and collapses the text into a single paragraph, but if you click through to the book details page you can see the nicely formatted summary.

packages/opds-web-client/src/components/BookCover.tsx
aslagle: Yes, this is for automatically generating book covers for books that don't have them, so the title and authors need to fit in a fixed size cover. The longer the text is the smaller it needs to be.

EdwinGuzman: This is great. We needed something like this for New Arrivals a while ago.

PR #222, opened 2018-09-20
Removing create-react-class dependency and converting component to a …

EdwinGuzman:
…class at the root level.

Fixes Jira [SIMPLY-625](https://jira.nypl.org/browse/SIMPLY-625). Converted a component into a class in order to remove the `create-react-class` dependency.

### DIFF
--- a/packages/opds-web-client/src/app.tsx
+++ b/packages/opds-web-client/src/app.tsx
@@ -11,8 +11,12 @@ import AuthPlugin from "./AuthPlugin";
 import "./stylesheets/app.scss";
 
 const OPDSCatalogRouterHandler = (config) => {
+  interface OPDSCatalogParams {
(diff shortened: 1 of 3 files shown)


packages/opds-web-client/src/app.tsx
```diff
@@ -10,6 +10,37 @@ import { State } from "./state";
 import AuthPlugin from "./AuthPlugin";
 import "./stylesheets/app.scss";
 
+const OPDSCatalogRouterHandler = (config) => {
+  interface OPDSCatalogProps {
+    params?: any;
```
aslagle: Isn’t this type just collection url and book url?

EdwinGuzman: Yes, updated it.

PR #225, opened 2018-11-21
No language search

EdwinGuzman:
Fixes [SIMPLY-1213](https://jira.nypl.org/browse/SIMPLY-1213). Adding `language=all` to query only if the `allLanguageSearch` prop is passed down and is true.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,3 @@
 language: node_js
 node_js:
   - "8"
-install: cd packages/opds-web-client && npm install
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
 language: node_js
-
 node_js:
   - "8"
+install: cd packages/opds-web-client && npm install
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,0 @@
-## Changelog
-
-### v0.0.3
-- Updated the `travis.yml` file to not build the `server` package in Travis-CI. This is to prevent a build failure caused by the `nypl-simplified-webpub-viewer` package.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,4 @@
+## Changelog
+
+### v0.0.3
+- Updated the `travis.yml` file to not build the `server` package in Travis-CI. This is to prevent a build failure caused by the `nypl-simplified-webpub-viewer` package.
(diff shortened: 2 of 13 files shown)


EdwinGuzman:
Updated the `travis.yml` file to specify the installation command. Now the server package will not be built on Travis, only the `opds-web-client` package will be built and the tests will still run.

EdwinGuzman:
Removed the `nypl-simplified-webpub-viewer` package, updated and added documentation, and bumped up version of the repo.

CHANGELOG.md
```diff
@@ -0,0 +1,4 @@
+## Changelog
```
aslagle: I don't think it makes sense to have a changelog for the top-level package - that's just a wrapper for the other ones to make it easier to run the server. if we're going to start a changelog i would make it for the opds-web-client package.

aslagle: which I see you also did :)

packages/opds-web-client/CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 ## Changelog
 
+### v0.1.26
+- Modified search url to include `language=all` to search all languages instead of just using the Accept-Language header from the browser.
```
aslagle: I would specify that this is an optional prop and won't be done by default.

packages/server/CHANGELOG.md
```diff
@@ -0,0 +1,4 @@
+## Changelog
```
aslagle: I wouldn't bother with this changelog either since this package is just for local testing and demo purposes and isn't published.

EdwinGuzman: Fine. I think it's useful for developers to know the server changes for local development though.

.travis.yml
```diff
@@ -1,4 +1,4 @@
 language: node_js
-
 node_js:
   - "8"
+install: cd packages/opds-web-client && npm install
```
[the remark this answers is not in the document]
EdwinGuzman: Yes, I can remove this but we don't need the server folder for unit tests that only run on Travis, so I think this is fine.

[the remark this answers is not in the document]
aslagle: I just think it's good to know if someone commits a change that breaks the server build, so we don't accidentally break the heroku app.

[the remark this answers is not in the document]
EdwinGuzman: Sounds good.

PR #227, opened 2019-04-12
Update webpack ts

EdwinGuzman:
This is a working _work in progress_.
Webpack, Typescript, unit tests, and configs have been updated. I will write details about the update to the code I did soon. I got the tests to pass and the builds to work, but when I try to import it into the circulation admin or registry admin, the apps can't seem to pick up the package. If this can be tested first and if it works then I'll continue with the PR.

Specifically, I don't think the parent app is picking up the build folder. It should be the `/lib` folder that gets picked up but this app still produces a `/dist` folder which I'm not sure is ever used.


packages/opds-web-client/package.json
```diff
@@ -13,9 +12,9 @@
   "scripts": {
     "lint": "tslint -c tslint.json src/*.ts src/*.tsx src/**/*.ts src/**/*.tsx src/**/**/*.ts src/**/**/*.tsx && sass-lint -c .sass-lint.yml -v -q",
     "test": "npm run lint && tsc && cp -r src/stylesheets lib && mocha --require lib/testHelper.js lib/__tests__/*.js lib/**/__tests__/*.js lib/**/**/__tests__/*.js",
+    "test-file": "npm run lint && tsc && cp -r src/stylesheets lib && mocha --require lib/testHelper.js",
     "dev": "webpack-dev-server --progress --history-api-fallback --hot --content-base build --port 8090 --config webpack.dev.config",
-    "prod": "NODE_ENV=production webpack -p --progress --display-modules --config webpack.prod.config && rm -rf dist/lib",
-    "prepublish": "typings install && rm -rf lib && tsc && cp -r src/stylesheets lib && npm run prod"
```
aslagle: Why is this gone instead of moved to `prepublishOnly`?

packages/opds-web-client/src/__tests__/DataFetcher-test.ts
```diff
@@ -11,7 +12,17 @@ describe("DataFetcher", () => {
     mockFetch = stub().returns(new Promise<any>((resolve, reject) => {
       resolve({ status: 200 });
     }));
-    fetch = mockFetch;
+    // mockFetch = new Promise<any>((resolve, reject) => {
+    //   resolve({ status: 200 });
+    // });
+    fetchMock
+      .mock("test-url", mockFetch)
```
aslagle: Can we get rid of the stub with this library? It looks like you can do `fetchMock.mock("test-url", 200)`

packages/opds-web-client/src/__tests__/DataFetcher-test.ts
```diff
@@ -107,21 +124,20 @@ describe("DataFetcher", () => {
...
           status: 401,
           text: () => new Promise((resolve, reject) => resolve("unauthorized"))
         });
       }));
-      fetch = mockFetch;
+      window.fetch = mockFetch;
 
       let fetcher = new DataFetcher();
-      fetcher.fetchOPDSData("test url").catch(err => {
+      fetcher.fetchOPDSData("test-url").catch(err => {
```
aslagle: I think you might have to return this for it to work? (looking at the mocha docs on returning a promise instead of using `done`)

packages/opds-web-client/src/authMiddleware.ts
```diff
@@ -122,6 +123,8 @@ export default (authPlugins: AuthPlugin[], pathFor: PathFor) => {
                       error,
                       attemptedProvider
                     ));
+                    // Because authProviders is empty for a test.
+                    resolve();
```
aslagle: I don't think we should add this for a test. What should happen if `authProviders` is actually empty?

packages/opds-web-client/src/components/BookDetails.tsx
```diff
@@ -61,9 +60,6 @@ export default class BookDetails<P extends BookDetailsProps> extends Book<P> {
                 { this.circulationInfo() }
               </div>
             </div>
-            <div className="right-column-links col-sm-3">
-              { this.rightColumnLinks() }
-            </div>
```
aslagle: Why did you take this out? The patron app overrides `rightColumnLinks` to put in the report a problem link.

aslagle: Maybe it needs a comment that it's intentionally empty.

packages/opds-web-client/src/components/BorrowButton.tsx
```diff
@@ -1,12 +1,12 @@
...
-export interface BorrowButtonProps extends React.HTMLProps<any> {
+export interface BorrowButtonProps extends React.HTMLProps<{}> {
   borrow: () => Promise<any>;
 }
 
-export default class BorrowButton extends React.Component<BorrowButtonProps, void> {
+export default class BorrowButton extends React.Component<BorrowButtonProps, {}> {
   render() {
-    let props = Object.assign({}, this.props);
+    let props = JSON.parse(JSON.stringify(this.props));
```
aslagle: The previous version seems safer in case the props have something that can't be stringified. Like `props.children`, if it's more than just a string.

packages/opds-web-client/src/components/CatalogLink.tsx
```diff
@@ -38,16 +38,16 @@ export default class CatalogLink extends React.Component<CatalogLinkProps, void>
...
 
     let location = this.context.pathFor(collectionUrl, bookUrl);
 
     // remove props that Link won't recognize
-    let props = Object.assign({}, this.props);
-    delete props["collectionUrl"];
-    delete props["bookUrl"];
+    let props = Object.assign({}, rest);
+    // delete props["collectionUrl"];
+    // delete props["bookUrl"];
```
aslagle: I think this was to prevent console warnings from having props that don't belong on a link?

packages/opds-web-client/src/components/DownloadButton.tsx
```diff
@@ -12,14 +12,14 @@ export interface DownloadButtonProps extends React.HTMLProps<any> {
...
-export default class DownloadButton extends React.Component<DownloadButtonProps, void> {
+export default class DownloadButton extends React.Component<DownloadButtonProps, {}> {
   constructor(props) {
     super(props);
     this.fulfill = this.fulfill.bind(this);
   }
 
   render() {
-    let props = Object.assign({}, this.props);
+    let props = JSON.parse(JSON.stringify(this.props));
```
aslagle: This one doesn't seem to have props that aren't stringifiable but I'm not sure what all the HTML props are.

packages/opds-web-client/src/components/Root.tsx
```diff
@@ -66,10 +66,10 @@ export interface RootProps extends StateProps {
...
   epubReaderUrlTemplate?: (epubUrl: string) => string;
   fetchPage?: (url: string) => Promise<CollectionData>;
-  Header?: new() => __React.Component<HeaderProps, any>;
-  Footer?: new() => __React.Component<FooterProps, any>;
-  CollectionContainer?: new() => __React.Component<CollectionContainerProps, any>;
-  BookDetailsContainer?: new() =>  __React.Component<BookDetailsContainerProps, any>;
+  Header?: React.ComponentClass<HeaderProps, {}>;
+  Footer?: React.ComponentClass<FooterProps, {}>;
+  CollectionContainer?: React.ComponentClass<CollectionContainerProps, {}>;
+  BookDetailsContainer?: React.ComponentClass<BookDetailsContainerProps, {}>;
```
aslagle: this is a lot nicer :)

packages/opds-web-client/src/components/Root.tsx
```diff
@@ -81,6 +81,8 @@ export interface RootProps extends StateProps {
   closeErrorAndHideAuthForm?: () => void;
   setPreference?: (key: string, value: string) => void;
   allLanguageSearch?: boolean;
+  router?: {};
+  pathFor?: () => {};
```
aslagle: how did these end up in the props?

packages/opds-web-client/src/components/Root.tsx
```diff
@@ -267,7 +269,6 @@ export class Root extends React.Component<RootProps, RootState> {
                     collection={this.collectionDataWithLoans()}
                     fetchPage={this.props.fetchPage}
                     isFetchingCollection={this.props.isFetchingCollection}
-                    isFetchingBook={this.props.isFetchingBook}
```
aslagle: what happened here?

packages/opds-web-client/src/components/__tests__/BorrowButton-test.tsx
```diff
@@ -7,6 +7,7 @@ import { shallow } from "enzyme";
 import BorrowButton from "../BorrowButton";
 import { ungroupedCollectionData } from "./collectionData";
 
+const stringify = (obj) => JSON.stringify(obj);
```
aslagle: I think you can just do `stringify = JSON.stringify;`

packages/opds-web-client/src/components/__tests__/BorrowButton-test.tsx
```diff
@@ -32,7 +32,7 @@ describe("BorrowButton", () => {
 
   it("shows button", () => {
     let button = wrapper.find("button");
-    expect(button.props().style).to.equal(style);
+    expect(stringify(button.props().style)).to.equal(stringify(style));
```
aslagle: alternately, you can probably use `deep.equal` instead of `equal`? That's probably better since things could end up in different orders in the strings.

packages/opds-web-client/webpack.common.js
```diff
@@ -0,0 +1,24 @@
...
+  module: {
+    rules: [
+      {
+        test: /\.(ttf|woff|eot|svg|png|woff2|gif|jpg)(\?[\s\S]+)?$/,
+        loader: 'url-loader?limit=100000'
+      }
+    ]
+  },
+  resolve: {
+    extensions: [".webpack.js", ".web.js", ".js", ".ts", ".tsx", ".scss"]
```
aslagle: I know you just moved this from another file but we could probably take out the first two of these.

aslagle:
The server package uses the `dist` folder - and it's not working in the heroku deployment.

packages/opds-web-client/src/components/BookDetails.tsx
EdwinGuzman: Ah yes, that would be better.

packages/opds-web-client/package.json
EdwinGuzman: I'm still trying to figure this part out. But now I realized where I could've gone wrong. I just realized that the `prepublish` command builds the lib folder.

packages/opds-web-client/src/__tests__/DataFetcher-test.ts
EdwinGuzman: Yes, I was still trying to figure out the full API of this mock package.

packages/opds-web-client/src/components/CatalogLink.tsx
EdwinGuzman: Got it. They did not come up when doing this because the `rest` only has the needed props.

packages/opds-web-client/src/components/BorrowButton.tsx
EdwinGuzman: I kept getting warnings about not being about to delete a property that is read-only. But the previous version is much safer and didn't know about the `props.children` case. Will change back.

packages/opds-web-client/src/components/Root.tsx
EdwinGuzman: Tslint complained. In some cases I wanted to just fix the issues and make sure everything was working before going back and fixing it.

packages/opds-web-client/src/components/Root.tsx
EdwinGuzman: This wasn't used in the app, but now I'm thinking it's for an inherited use-case?

EdwinGuzman:
So both the `dist` and `lib` folder are build in the process: `lib` for the npm module and `dist` just for the demo app, right? I'll make a note of that. Thanks!

packages/opds-web-client/src/components/CatalogLink.tsx
aslagle: Maybe you could use this approach to get rid of the `JSON.stringify`s in other places.

packages/opds-web-client/src/components/Root.tsx
aslagle: I'm not sure about this one - it seems like it should be needed though, to show the loading message if you go directly to a book with no collection.

aslagle:
I believe `dist` is also published to npm and could be used, if you just want to include the library in an HTML page and you're not building another app around it. (which is similar to what the demo server in here does I guess)

EdwinGuzman:
New deployment works. I still need to figure out why using `npm link` or pointing to this branch in `package.json` has not been working for me. Maybe now it'll work, and if it does I'll go back and clean up the rest.

packages/opds-web-client/src/__tests__/DataFetcher-test.ts
EdwinGuzman: It times out when returning the promise because it seems to be expecting it to resolve and not reject.

packages/opds-web-client/src/authMiddleware.ts
EdwinGuzman: I'm not too sure what this part is testing but it seems like it will just end here without a `resolve` or a `reject`.

packages/opds-web-client/src/components/DownloadButton.tsx
EdwinGuzman: If I change it to destructure and use `...rest` and `Object.assign`, I get errors that the props variable is not compatible with the anchor element's prop interface because it is missing properties. I don't know how to fix that just yet.

packages/opds-web-client/src/__tests__/DataFetcher-test.ts
aslagle: In that case I think you might have to use `done` - otherwise the test will return before the promise rejects. Can you make it fail if you change something inside the catch?

packages/opds-web-client/src/authMiddleware.ts
aslagle: I still don't understand this. There's a test that breaks if you take this out?

packages/opds-web-client/src/components/CatalogLink.tsx
```diff
@@ -26,7 +26,7 @@ export default class CatalogLink extends React.Component<CatalogLinkProps, void>
   // and https://github.com/reactjs/react-router/blob/master/modules/PropTypes.js
   getChildContext() {
     let noop = () => {};
-    let router = Object.assign({}, this.context.router, {
+    let router: any = Object.assign({}, this.context.router, {
```
aslagle: You could probably use a type from react router?

packages/opds-web-client/src/reducers/index.ts
```diff
@@ -6,7 +6,7 @@ import loans from "./loans";
 import preferences from "./preferences";
 import { State } from "../state";
 
-const reducers = combineReducers<State>({
+const reducers: any = combineReducers<State>({
```
aslagle: looks like this should be `Reducer<State>`

packages/opds-web-client/src/components/BorrowButton.tsx
EdwinGuzman: I'm having trouble using `Object.assign` because using it makes Typescript think that it's missing ~250 props that belong on an anchor element (like `focus`, `aria-selected`, etc). I think it's fine to use `JSON.parse(JSON.stringify(...))` here because the only props that are not needed are `borrow` and `children`.

This may also help remove any props that are not needed:
```
  const { borrow, children, ...rest } = this.props;
  let props = JSON.parse(JSON.stringify(rest));
```

But doing:
```
  const { borrow, children, ...rest } = this.props;
  let props = Object.assign({}, rest);
```

Still returns
```
...
            Type '{}' is missing the following properties from type 'HTMLButtonElement': autofocus, disabled, form, formAction, and 255 
```

This works but defeats the purpose: `let props = Object.assign<any, any>({}, this.props);`

EdwinGuzman: What do you think of doing `let props = Object.assign<HTMLButtonElement, any>({} as HTMLButtonElement, this.props);`?

aslagle: If you use the `...rest` version, I don't think you still need the `Object.assign` at all? I think it was making a copy so it could delete some props without deleting them from the original, but `rest` already accomplishes that.

EdwinGuzman: Just using `rest` without copying also throws the same error.

packages/opds-web-client/src/__tests__/DataFetcher-test.ts
EdwinGuzman: Fixed the test and the function.

packages/opds-web-client/src/authMiddleware.ts
EdwinGuzman: These tests https://github.com/NYPL-Simplified/opds-web-client/blob/ac847e1c6072f7da8e6effc2129d99a156ed59a1/packages/opds-web-client/src/__tests__/authMiddleware-test.ts#L114 get hung up because it's waiting to be resolved to go to the `.then()` but if you remove this `resolve`, then the tests just time out. It's not just the test in the link but that test and every single test after it.

packages/opds-web-client/src/components/BorrowButton.tsx
aslagle: It seems to work if you do `const { ref, borrow, children, ...rest } = this.props;` - I think the `ref` prop causes the error.

packages/opds-web-client/src/__tests__/authMiddleware-test.ts
```diff
@@ -36,10 +20,11 @@ describe("authMiddleware", () => {
   beforeEach(() => {
     dataFetcher = new DataFetcher();
     dataFetcher.clearAuthCredentials();
-    hideAuthFormStub = stub();
-    clearAuthCredentialsStub = stub();
-    showAuthFormStub = stub();
-    actionCreatorStub = stub(ActionCreator, "default", MockActionCreator);
+    // clearAuthCredentialsStub = stub();
+    // showAuthFormStub = stub();
```
aslagle: you can remove these.

packages/opds-web-client/src/authMiddleware.ts
aslagle: Okay, I think I get it now. We were leaving an unresolved promise after the authentication had failed - we caught the error from a different promise, but didn't resolve or reject this promise. I think it's okay to resolve since the problem was handled, but I would take out the comment about the test. I think leaving it unresolved also shows errors in the browser console.

packages/opds-web-client/src/__tests__/authMiddleware-test.ts
EdwinGuzman: This is weird. I did remove them and they're not showing up anymore locally or in this file but it is for this PR change view.

packages/opds-web-client/src/components/BorrowButton.tsx
EdwinGuzman: Yes! Thanks, it is `ref` that causes the error. Will update the rest.

EdwinGuzman:
I am not sure what else this is missing. Can you give it another look?

Once this is merged and published, I'll bring it into the Registry Admin PR https://github.com/NYPL-Simplified/registry_admin/pull/15 and then make a PR for Circulation Web.

packages/opds-web-client/package.json
```diff
@@ -12,10 +12,11 @@
   "license": "Apache-2.0",
   "scripts": {
     "lint": "tslint -c tslint.json src/*.ts src/*.tsx src/**/*.ts src/**/*.tsx src/**/**/*.ts src/**/**/*.tsx && sass-lint -c .sass-lint.yml -v -q",
-    "test": "npm run lint && tsc && cp -r src/stylesheets lib && mocha --require lib/testHelper.js lib/__tests__/*.js lib/**/__tests__/*.js lib/**/**/__tests__/*.js",
+    "test": "rm -rf lib npm run lint && tsc && cp -r src/stylesheets lib && mocha --require lib/testHelper.js lib/__tests__/*.js lib/**/__tests__/*.js lib/**/**/__tests__/*.js",
+    "test-file": "npm run lint && tsc && cp -r src/stylesheets lib && mocha --require lib/testHelper.js",
     "dev": "webpack-dev-server --progress --history-api-fallback --hot --content-base build --port 8090 --config webpack.dev.config",
-    "prod": "NODE_ENV=production webpack -p --progress --display-modules --config webpack.prod.config && rm -rf dist/lib",
```
aslagle: Can you confirm that dist/lib is no longer being generated? I don't remember why this was here but don't want to be publishing it if it is still created.

aslagle: Maybe the `rm -rf lib` in prepublish handles it.

packages/opds-web-client/src/components/BorrowButton.tsx
```diff
@@ -1,15 +1,13 @@
...
 
-export default class BorrowButton extends React.Component<BorrowButtonProps, void> {
+export default class BorrowButton extends React.Component<BorrowButtonProps, {}> {
   render() {
-    let props = Object.assign({}, this.props);
-    delete props["book"];
-    delete props["borrow"];
-    delete props["key"];
+    const { ref, borrow, children, ...rest } = this.props;
+    let props = Object.assign({}, rest);
```
aslagle: The spread operator already makes a new object so I don't think there's any need for the Object.assign.

packages/opds-web-client/src/components/Root.tsx
aslagle: I'd still like to investigate this.

packages/opds-web-client/package.json
EdwinGuzman: The command in prepublish was added because I encountered overriding issues (maybe just my laptop?) but I added back the command to delete `dist/lib`. That folder will be gone now. Thanks for that catch.

packages/opds-web-client/src/components/BorrowButton.tsx
EdwinGuzman: Gotcha, will update the three cases where I used this technique.

packages/opds-web-client/src/components/Root.tsx
EdwinGuzman: Okay, I see what was wrong. The `contextTypes` were being passed the `RootProps` - `React.ValidationMap<RootProps>`. It should be the `NavigateContext` - `React.ValidationMap<NavigateContext>`.

EdwinGuzman:
Alright, updated the type for the `Root`'s context and removed some unused imports.

aslagle:
When I go to NYPL's feed in the Heroku app and click "My Books", it doesn't work. It shows the main feed again instead of the login popup. I think it might be an issue with the auth middleware.

packages/opds-web-client/src/authMiddleware.ts
```diff
@@ -9,7 +9,7 @@ import { AuthCallback, AuthProvider, AuthMethod, PathFor } from "./interfaces";
     See Redux Middleware docs:
     http://redux.js.org/docs/advanced/Middleware.html */
 
-export default (authPlugins: AuthPlugin[], pathFor: PathFor) => {
+export default (authPlugins: AuthPlugin[], pathFor: PathFor, test: boolean = false) => {
```
EdwinGuzman: @aslagle I don't currently see another way of solving this except by doing this technique which is used often in the server code. The tests run and the login form should work now.

packages/opds-web-client/src/components/Book.tsx
```diff
@@ -144,10 +144,11 @@ export default class Book<P extends BookProps> extends React.Component<P, {}> {
 
     if (this.isOpenAccess()) {
       if (this.props.epubReaderUrlTemplate) {
+        let index = 0;
         for (const link of this.props.book.openAccessLinks) {
           if (link.type === "application/epub+zip") {
             links.push(
-              <span key={link.url}>
+              <span key={`${link.url}-${index}`}>
```
EdwinGuzman: Added these to remove the console warnings about react flatten children having the same key in a render loop.

packages/opds-web-client/src/authMiddleware.ts
aslagle: I think I figured it out. The resolve that's supposed to get called is on line 73, in the callback that's passed to `showAuthMiddleware`. But `showAuthMiddleware` is a stub, so you have to use `showAuthFormStub.callsArg(0);` to make it resolve.

packages/opds-web-client/src/__tests__/authMiddleware-test.ts
```diff
@@ -2,30 +2,15 @@ import { expect } from "chai";
...
-  hideAuthForm() {
-    return hideAuthFormStub();
-  }
-
-  clearAuthCredentials() {
-    return clearAuthCredentialsStub();
-  }
-
-  showAuthForm(callback, cancel, authProviders, title, error, attemptedProvider) {
-    callback();
```
aslagle: And here's how it was getting resolved before.

packages/opds-web-client/src/authMiddleware.ts
EdwinGuzman: That makes sense. Thanks!

EdwinGuzman:
Great, thanks!

PR #228, opened 2019-05-16
Redux update

EdwinGuzman:
Not updating to Redux 4 since that depends on React 16. There are no significant updates but this was also helpful for me to get better acquainted with the actions and reducers in this app. I tried to use `fetch-mock` for some of fetch tests but it doesn't support `Blob` so I kept the previous mocking class.


aslagle:
shouldn't there be an update to package-lock?

packages/opds-web-client/src/DataFetcher.ts
```diff
@@ -70,8 +70,15 @@ export default class DataFetcher {
...
           });
-        })
-      }).catch(reject);
+        });
+      }).catch(error => {
+        reject({
+          status: error.status,
+          response: error.message,
+          url: url,
+          headers: error.headers
```
aslagle: I'm not sure the error will have all these properties - for example, https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch#Checking_that_the_fetch_was_successful says it can return a `TypeError`, which would only have `message`. Weird that Typescript doesn't complain.

EdwinGuzman: Yeah, the error value itself is the message string so there's no object.

EdwinGuzman:
Added `package-lock.json`.

PR #229, opened 2019-05-20
React16

EdwinGuzman:
Fixes [SIMPLY-1965](https://jira.nypl.org/browse/SIMPLY-1965) by updating syntax and other related packages that are needed for React 16. Also updated tests to use the updated Enzyme API.


packages/opds-web-client/src/__tests__/setupEnyzme.ts
```diff
@@ -0,0 +1,4 @@
+import { configure } from 'enzyme';
+import * as Adapter from 'enzyme-adapter-react-16';
+
+configure({ adapter: new Adapter() });
```
aslagle: How does this work with nothing referencing this file? Could it be combined with src/testHelper?

packages/opds-web-client/src/components/BasicAuthForm.tsx
```diff
@@ -60,14 +60,20 @@ export default class BasicAuthForm extends React.Component<BasicAuthFormProps, B
...
+    let password = (this.refs["password"] as any).value;
 
     if (!login) {
       this.setState({
         error: `${this.loginLabel()} is required`
       });
       return false;
+    } else if (!password) {
+      this.setState({
+        error: `${this.loginLabel()} is required`
```
aslagle: There are some libraries where a password is not required. I think there's something in the auth document that indicates that.

packages/opds-web-client/src/components/__tests__/Collection-test.tsx
```diff
@@ -112,30 +128,37 @@ describe("Collection", () => {
...
       );
-      let viewToggleButtons = wrapper.find(".view-toggle button");
-      expect(viewToggleButtons.length).to.equal(2);
+      const getElements = () => {
+        let viewToggleButtons = wrapper.find(".view-toggle button");
+
+        return [
+          viewToggleButtons,
+          viewToggleButtons.at(0), // gridButton
+          viewToggleButtons.at(1), // listButton
```
aslagle: You could make this an object and then the keys could replace these comments.

packages/opds-web-client/src/components/__tests__/UrlForm-test.tsx
```diff
@@ -33,9 +33,9 @@ describe("UrlForm", () => {
     );
 
     let form = wrapper.find("form");
-    let input = wrapper.find("input").get(0) as any;
-
-    input.value = "some url";
+    
+    // set the value for the input
+    wrapper.setProps({ collectionUrl: "some url" });
```
aslagle: It's good to test using the props, but it would still be good to also check that something typed in works.

packages/opds-web-client/src/__tests__/setupEnyzme.ts
EdwinGuzman: Yeah, it does work. Will update.

packages/opds-web-client/src/components/BasicAuthForm.tsx
EdwinGuzman: Gotcha. I'll write a comment for that because it confused me.

packages/opds-web-client/src/components/__tests__/UrlForm-test.tsx
EdwinGuzman: Yes, I figured out how to do so later on and forgot to update this one.

packages/opds-web-client/src/components/__tests__/Collection-test.tsx
EdwinGuzman: I initially did that and ended up with `let { viewToggleButtons, gridButton, listButton, books } = getElements();` but in this test when `wrapper.setProps({ preferences });` is called, the components do not update. If you do `wrapper.find(...)` again and refetch the elements, the correct values are there. So I did the object version but couldn't redeclare the variables.

It'd be better if it wasn't an issue to get the element once and get the updated value from it without having to look it up again. Not sure why it's happening for this case.

aslagle: Yeah, that would be better, but seems like it must've been something that changed in enzyme. Anyway, I was just saying you could have this function return an object instead of a list, but you would still need the function.

EdwinGuzman: I can't call `let { viewToggleButtons, gridButton, listButton, books } = getElements();` repeatedly though and that's why I made it an array. I'll see if I can figure out something else.

aslagle: Oh interesting, I didn't realize you couldn't do that. Apparently you can put parentheses around it for the subsequent times though - `({ viewToggleButtons, gridButton, listButton, books } = getElements());`

EdwinGuzman: It seems like very odd syntax to me but 👍 .

PR #230, opened 2019-06-02
Added subtitle and fixed publication date on book details page.

aslagle:
This uses https://github.com/NYPL-Simplified/opds-feed-parser/pull/44, which needs to be published before I can deploy a demo or get tests running here.

It fixes a couple things that are confusing in the admin interface - if you add a subtitle in the edit tab, it does not show up on the details tab. And if you add a publication date, it will not have a time, so due to time zones the details tab would display one day before the date you added.

### DIFF
--- a/packages/opds-web-client/src/__tests__/OPDSDataAdaptor-test.ts
+++ b/packages/opds-web-client/src/__tests__/OPDSDataAdaptor-test.ts
@@ -52,10 +52,11 @@ describe("OPDSDataAdapter", () => {
       title: "The Mayan Secrets",
       authors: [factory.contributor({name: "Clive Cussler"}), factory.contributor({name: "Thomas Perry"})],
       contributors: [factory.contributor({name: "contributor"})],
+      subtitle: "A Sam and Remi Fargo Adventure",
       summary: factory.summary({content: "&lt;b&gt;Sam and Remi Fargo race for treasure&#8212;and survival&#8212;in this lightning-paced new adventure from #1&lt;i&gt; New York Times&lt;/i&gt; bestselling author Clive Cussler.&lt;/b&gt;&lt;br /&gt;&lt;br /&gt;Husband-and-wife team Sam and Remi Fargo are in Mexico when they come upon a remarkable discovery&#8212;the mummified remainsof a man clutching an ancient sealed pot. Within the pot is a Mayan book larger than any known before.&lt;br /&gt;&lt;br /&gt;The book contains astonishing information about the Mayans, their cities, and about mankind itself. The secrets are so powerful that some people would do anything to possess them&#8212;as the Fargos are about to find out. Many men and women are going to die for that book.<script>alert('danger!');</script>"}),
       categories: [factory.category({label: "label"}), factory.category({term: "no label"}), factory.category({label: "label 2"})],
       links: [largeImageLink, thumbImageLink, openAccessLink, borrowLink, fulfillmentLink, collectionLink],
-      issued: "2014-06-08T22:45:58Z",
+      issued: "2014-06-08",
       publisher: "Fake Publisher",
       series: {
         name: "Fake Series",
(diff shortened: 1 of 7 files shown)


packages/opds-web-client/src/components/BookDetails.tsx
```diff
@@ -19,6 +19,11 @@ export default class BookDetails<P extends BookDetailsProps> extends Book<P> {
           </div>
           <div className="header">
             <h1 className="title">{this.props.book.title}</h1>
+            {
+              this.props.book.subtitle ?
+              <h3 className="subtitle">{ this.props.book.subtitle }</h3> :
+              ""
```
EdwinGuzman: Rendering null or using `&&` syntax may be better than an empty string.

packages/opds-web-client/src/__tests__/OPDSDataAdaptor-test.ts
```diff
@@ -52,10 +52,11 @@ describe("OPDSDataAdapter", () => {
...
+      subtitle: "A Sam and Remi Fargo Adventure",
       summary: factory.summary({content: "&lt;b&gt;Sam and Remi Fargo race for treasure&#8212;and survival&#8212;in this lightning-paced new adventure from #1&lt;i&gt; New York Times&lt;/i&gt; bestselling author Clive Cussler.&lt;/b&gt;&lt;br /&gt;&lt;br /&gt;Husband-and-wife team Sam and Remi Fargo are in Mexico when they come upon a remarkable discovery&#8212;the mummified remainsof a man clutching an ancient sealed pot. Within the pot is a Mayan book larger than any known before.&lt;br /&gt;&lt;br /&gt;The book contains astonishing information about the Mayans, their cities, and about mankind itself. The secrets are so powerful that some people would do anything to possess them&#8212;as the Fargos are about to find out. Many men and women are going to die for that book.<script>alert('danger!');</script>"}),
       categories: [factory.category({label: "label"}), factory.category({term: "no label"}), factory.category({label: "label 2"})],
       links: [largeImageLink, thumbImageLink, openAccessLink, borrowLink, fulfillmentLink, collectionLink],
-      issued: "2014-06-08T22:45:58Z",
```
EdwinGuzman: Does this work for both syntaxes? Or will the server now return the latter?

packages/opds-web-client/src/components/BookDetails.tsx
```diff
@@ -19,6 +19,11 @@ export default class BookDetails<P extends BookDetailsProps> extends Book<P> {
           </div>
           <div className="header">
             <h1 className="title">{this.props.book.title}</h1>
+            {
+              this.props.book.subtitle ?
+              <h3 className="subtitle">{ this.props.book.subtitle }</h3> :
```
EdwinGuzman: I see that the current heading structure is
```
H1 // title
H3 // if there is a subtitle
H2 // authors
```

It'd be better if the H2 came before the H3. I'm not sure this should be an h3 or authors an h2 but not part of the PR (and I think technically only the homepage of a site should have an h1).

packages/opds-web-client/src/__tests__/OPDSDataAdaptor-test.ts
aslagle: Yeah, it works with either. The previous syntax is already UTC, so the change does nothing.

packages/opds-web-client/src/components/BookDetails.tsx
aslagle: Hmm, I think authors are more important than subtitle, but it seems like it would look weird to put the subtitle below the authors.

aslagle: This is also a problem with series.

EdwinGuzman: I agree it would look weird. This should be a separate conversation.

killerpuppytails: Hey, this is Risa - would you like me to bring this to our accessibility/UX team for recommendations, or discuss on your own?

aslagle: Yeah I think that would be useful.

EdwinGuzman: Yes, @killerpuppytails can you make a ticket so Willa and I can work on it? I'll then loop you in @aslagle .

killerpuppytails: Sure thing!

killerpuppytails: SIMPLY-2022 created and assigned to you @EdwinGuzman

PR #231, opened 2019-06-05
Updating the headings on the book detail page.

EdwinGuzman:
Minor updates for https://jira.nypl.org/browse/SIMPLY-2023. From a conversation with Willa, the current headings don't define any sections so they don't need to be headings. The summary section does need a heading so I added one there. There are also a few CSS updates and the book details is now a list.

### DIFF
--- a/packages/opds-web-client/src/components/BookDetails.tsx
+++ b/packages/opds-web-client/src/components/BookDetails.tsx
@@ -44,7 +44,7 @@ export default class BookDetails<P extends BookDetailsProps> extends Book<P> {
               ) }
               {
                 medium && (
-                  <li className="item-icon-container">Format: {this.getMediumSVG(medium)}</li>
+                  <li className="item-icon-container">{this.getMediumSVG(medium)}</li>
                 )
               }
             </ul>
--- a/packages/opds-web-client/src/components/BookDetails.tsx
+++ b/packages/opds-web-client/src/components/BookDetails.tsx
@@ -17,43 +17,43 @@ export default class BookDetails<P extends BookDetailsProps> extends Book<P> {
           <div className="cover">
             <BookCover book={this.props.book} />
           </div>
-          <div className="header">
+          <div className="details">
             <h1 className="title">{this.props.book.title}</h1>
             {
               this.props.book.subtitle &&
-              <h3 className="subtitle">{ this.props.book.subtitle }</h3>
+              <p className="subtitle">{ this.props.book.subtitle }</p>
             }
             {
               this.props.book.series && this.props.book.series.name &&
-              <h3 className="series">{ this.props.book.series.name }</h3>
+              <p className="series">{ this.props.book.series.name }</p>
             }
             {
-              this.props.book.authors && this.props.book.authors.length &&
-              <h2 className="authors">By {this.props.book.authors.join(", ")}</h2>
+              this.props.book.authors && this.props.book.authors.length > 0 &&
+              <p className="authors">By {this.props.book.authors.join(", ")}</p>
             }
             {
-              this.props.book.contributors && this.props.book.contributors.length &&
-              <h2 className="contributors">Contributors: {this.props.book.contributors.join(", ")}</h2>
+              this.props.book.contributors && this.props.book.contributors.length > 0 &&
+              <p className="contributors">Contributors: {this.props.book.contributors.join(", ")}</p>
             }
-            <div className="fields" lang="en">
+            <ul className="fields" lang="en">
               { fields.map(field =>
-                field.value ? <div className={field.name.toLowerCase().replace(" ", "-")} key={field.name}>{field.name}: {field.value}</div> : null
+                field.value &&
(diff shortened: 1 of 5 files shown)


packages/opds-web-client/src/components/BookDetails.tsx
```diff
@@ -17,43 +17,43 @@ export default class BookDetails<P extends BookDetailsProps> extends Book<P> {
...
             }
             {
               this.props.book.series && this.props.book.series.name &&
-              <h3 className="series">{ this.props.book.series.name }</h3>
+              <p className="series">{ this.props.book.series.name }</p>
             }
             {
-              this.props.book.authors && this.props.book.authors.length &&
-              <h2 className="authors">By {this.props.book.authors.join(", ")}</h2>
+              this.props.book.authors && this.props.book.authors.length > 0 &&
```
EdwinGuzman: Updated this because `0` was printed out before.

packages/opds-web-client/src/components/BookDetails.tsx
```diff
@@ -17,43 +17,43 @@ export default class BookDetails<P extends BookDetailsProps> extends Book<P> {
...
+              <p className="authors">By {this.props.book.authors.join(", ")}</p>
             }
             {
-              this.props.book.contributors && this.props.book.contributors.length &&
-              <h2 className="contributors">Contributors: {this.props.book.contributors.join(", ")}</h2>
+              this.props.book.contributors && this.props.book.contributors.length > 0 &&
+              <p className="contributors">Contributors: {this.props.book.contributors.join(", ")}</p>
             }
-            <div className="fields" lang="en">
+            <ul className="fields" lang="en">
```
aslagle: This being a list seems a bit weird since these are a bunch of unrelated items, but if you all think it makes sense that's fine.

packages/opds-web-client/src/components/BookDetails.tsx
```diff
@@ -17,43 +17,43 @@ export default class BookDetails<P extends BookDetailsProps> extends Book<P> {
...
               ) }
-            </div>
-            {
-              medium && (
-                <div className="item-icon-container">{this.getMediumSVG(medium)}</div>
-              )
-            }
+              {
+                medium && (
+                  <li className="item-icon-container">Format: {this.getMediumSVG(medium)}</li>
```
aslagle: What was the reasoning for this change?

aslagle: Also, it's inconsistent with the list view now.

packages/opds-web-client/src/components/BookDetails.tsx
EdwinGuzman: They were all grouped before and it's a list of the book's detail, so not exactly unrelated.

packages/opds-web-client/src/components/BookDetails.tsx
EdwinGuzman: Where in the list view?
It's to be more explicit about why the label is there for.

aslagle: When you're on the scrolling list of books and you change to list view, it still just has the icon and eBook. It also has it higher up, so it already wasn't consistent.

Personally I didn't think this was unclear... And I'm not sure "Format" would be more clear.

EdwinGuzman: Do you mean on this page: 
Screen Shot 2019-06-05 at 3 22 37 PM

I don't see the format now (on this or the opds demo app).

aslagle: Yeah, that one, but using a wider window. It hides some of it if your window is small.

EdwinGuzman: I think it's very clear for us since we know what we are seeing since we work on it, but I think it's better to not make users think too much about the data they are seeing. It is pretty clear already what the icon and label represent, but there could be some that find it harder to know what it means without a definition.

aslagle: I think it could be confusing if you don't know what the possible formats are, but adding the word "Format" doesn't explain that anyway.

EdwinGuzman: Ah, I see it. Since this PR is just for the book detail page, I rather update that grid/list page in a different PR.

EdwinGuzman: I think it'd be good to group them together if possible but if "format" doesn't work, would another word work better?

aslagle: "Format" does seem like the appropriate word but I don't feel like it adds anything here. I guess I don't understand the problem.

EdwinGuzman: It's not a problem per se, but we have a definition list and the item's format label is there but not the term. It's not an actual `<dl>` (which it could be) but adding format there makes it consistent with the information displayed above.

PR #232, opened 2019-06-26
Redux update

aslagle:
This updates Redux to version 4, and also updates a few other packages. It turned out to be pretty straightforward - there were only a few changes and some of them were just fixing types that were wrong before.


packages/opds-web-client/src/components/Root.tsx
```diff
@@ -409,7 +409,7 @@ export class Root extends React.Component<RootProps, RootState> {
   }
 }
 
-let connectOptions = { withRef: true, pure: false };
+let connectOptions = { pure: false };
```
aslagle: @EdwinGuzman I think you understand this better than I do from your previous updates. I feel like this may have already stopped working in the previous update but it didn't get caught before because we didn't have the new Redux typings. You apparently can use `forwardRef: true` here instead, but I'm not sure if we need that.

packages/opds-web-client/src/store.ts
```diff
@@ -23,7 +23,7 @@ export default function buildStore(initialState?: State, authPlugins?: AuthPlugi
   if (persistState) {
     composeArgs.push(persistState("preferences"));
   }
-  return createStore<State>(
+  return createStore(
```
aslagle: This was asking for 4 parameters instead of 1 with the new typings, but the documentation doesn't use any and it seems to work fine without them.

packages/opds-web-client/src/components/Root.tsx
EdwinGuzman: I see. Adriana is working on an update for the form in the admins and it doesn't seem to need the `forwardRef` at all but this is good to know.

aslagle:
Yeah, there aren't a lot of changes. I think we could add more type checking around actions now if we want to, but I didn't do it here. I mostly wanted to do this update so we could follow the latest documentation.

PR #238, opened 2019-10-30
Data fetcher adapter

EdwinGuzman:
Resolves [SIMPLY-623](https://jira.nypl.org/browse/SIMPLY-623).

Rejecting `fetchOPDSData` if no adapter is configured in `DataFetcher`. Minor update to the error message that gets rendered to include the response.
Screen Shot 2019-10-30 at 3 00 37 PM


EdwinGuzman:
Yes, the least they can do with the information is report it, but if a developer looks at it they will have better knowledge about it (if they don't look at the console).

PR #240, opened 2019-11-06
Adding react-axe and a script to run it locally.

EdwinGuzman:
Resolves [SIMPLY-2375](https://jira.nypl.org/browse/SIMPLY-2375).
Adding `react-axe` and a script to run it. It should only be used for testing/fixing accessibility bugs but not during active development because it is resource intensive.

### DIFF
--- a/packages/opds-web-client/src/DataFetcher.ts
+++ b/packages/opds-web-client/src/DataFetcher.ts
@@ -37,10 +37,6 @@ export default class DataFetcher {
     this.proxyUrl = config.proxyUrl;
     this.adapter = config.adapter;
     this.authKey = "authCredentials";
-
-    if (!this.adapter) {
-      console.error("No adapter has been configured.");
-    }
   }
 
   fetchOPDSData(url: string) {
--- a/packages/opds-web-client/src/app.tsx
+++ b/packages/opds-web-client/src/app.tsx
@@ -66,6 +66,15 @@ class OPDSWebClient {
 
   render() {
     let RouteHandler = this.RouteHandler;
+
+    // `react-axe` should only run in development and testing mode.
+    // Running this is resource intensive and should only be used to test
+    // for accessibility and not while active development.
+    if (process.env.TEST_AXE === "true") {
+      let axe = require("react-axe");
+      axe(React, ReactDOM, 1000);
+    }
+
     ReactDOM.render(
       <Router history={browserHistory}>
         <Route path={this.pathPattern} component={RouteHandler} />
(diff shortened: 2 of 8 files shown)


packages/opds-web-client/src/DataFetcher.ts
```diff
@@ -37,10 +37,6 @@ export default class DataFetcher {
     this.proxyUrl = config.proxyUrl;
     this.adapter = config.adapter;
     this.authKey = "authCredentials";
-
-    if (!this.adapter) {
```
EdwinGuzman: @aslagle removing this because the app re-renders and runs this code many times. Then this gets outputted in the console many times (falsely) until the adapter is set. I think there's a larger problem with how some parts of the app re-render more times than it should.

packages/opds-web-client/src/app.tsx
```diff
@@ -66,6 +66,15 @@ class OPDSWebClient {
 
   render() {
     let RouteHandler = this.RouteHandler;
+
+    // `react-axe` should only run in development and testing mode.
+    // Running this is resource intensive and should only be used to test
+    // for accessibility and not while active development.
```
adriana-alter: Same comment as in registry_admin PR: "while actively developing" or "during active development".

packages/opds-web-client/src/DataFetcher.ts
aslagle: Yeah, if it was up to me I would definitely prioritize performance and updating the app to follow react/redux best practices over all the UX stuff.

packages/opds-web-client/src/app.tsx
```diff
@@ -66,6 +66,15 @@ class OPDSWebClient {
 
   render() {
     let RouteHandler = this.RouteHandler;
+
+    // `react-axe` should only run in development and testing mode.
+    // Running this is resource intensive and should only be used to test
+    // for accessibility and not during active development.
+    if (process.env.TEST_AXE === "true") {
```
aslagle: Does this part of the code not run in the browser?

aslagle: Oh I see, you define it in webpack

aslagle: `process.env` is a little confusing because that's normally used for environment variables on the server

EdwinGuzman: Yes, that is true but that's the convention that is usually used, even in files that will be used in the browser only.

EdwinGuzman:
@aslagle  I didn't mentioned you in this ticket, sorry about that. Here's some investigation I did regarding accessibility tools: https://jira.nypl.org/browse/SIMPLY-2366

It's agreed we'd all prefer a tool that can run automatically, especially in travis but for accessibility and UX it's slightly more complicated.

PR #241, opened 2019-11-06
Adding a loadCollection action to load any loans collection data…

EdwinGuzman:
…which was already fetch to improve performance.

Resolves [SIMPLY-613](https://jira.nypl.org/browse/SIMPLY-613). Adding this update to the existing unpublished version 0.2.9.

When a loans collection is fetched, it will be saved in the store. Then, when attempting to fetch the same loans url, instead of making a new request, it will load the loans collection. There is some speed improvements when going back and forth between the catalog and the "My Books" tab.

Two things concern me:
1) The catalog breadcrumbs gets lost so I am using the current (catalog collection) breadcrumb location url and adding it to the loans collection object. Otherwise, it would be as if you go directly to the "My Books" (loans) page and there is no way to go to the catalog.
2) If you borrow a book, the books in the loans collection will correctly update so there is no need to make a new request. The only issue is that it takes about 1-2 seconds for that process to happen right after clicking "borrow" on a book. So a user decides to borrow a book and then right away go to the "My Books" loans tab, the book won't appear. Not sure how often that will happen but if the user then goes to the catalog and back again, the book will reappear. This isn't an issue if a user clicks on "borrow" and then takes their time to go to the loans page.

The best way to test would be by signing into your library and borrow some books.

### DIFF
--- a/packages/opds-web-client/src/components/Book.tsx
+++ b/packages/opds-web-client/src/components/Book.tsx
@@ -208,7 +208,7 @@ export default class Book<P extends BookProps> extends React.Component<P, {}> {
       );
       links.push(
         <BorrowButton
-          key={this.props.book.borrowUrl}
+          key={`${this.props.book.borrowUrl}-borrow-button`}
           className="btn btn-default borrowed-button"
           disabled={true}
           borrow={this.borrow}>
--- a/packages/opds-web-client/src/components/__tests__/mergeRootProps-test.ts
+++ b/packages/opds-web-client/src/components/__tests__/mergeRootProps-test.ts
@@ -5,8 +5,8 @@ import ActionCreator from "../../actions";
 
 // synchronous actions for simple testing
 // of createFetchCollectionAndBook
-let fetchCollectionStub = stub().returns(new Promise((resolve, reject) => { resolve({}); }));
-let fetchBookStub = stub().returns(new Promise((resolve, reject) => { resolve({}); }));
+let fetchCollectionStub = stub().returns(Promise.resolve({}));
+let fetchBookStub = stub().returns(Promise.resolve({}));
 
 import { mergeRootProps, findBookInCollection, createFetchCollectionAndBook } from "../mergeRootProps";
 import { groupedCollectionData, ungroupedCollectionData } from "./collectionData";
@@ -57,7 +57,7 @@ describe("createFetchCollectionAndBook", () => {
     actionBookCollectionStub.restore();
   });
 
-  it("returns fetch function that uses the provided dispatch", (done) => {
+  it("returns fetch function that uses the provided dispatch", async () => {
     let fetchCollectionAndBook = createFetchCollectionAndBook(dispatch);
     fetchCollectionAndBook(collectionUrl, bookUrl)
       .then(({ collectionData, bookData }) => {
@@ -66,7 +66,6 @@ describe("createFetchCollectionAndBook", () => {
         expect(dispatch.callCount).to.equal(2);
         expect(dispatch.args[0][0]).to.equal(fetchCollectionStub);
         expect(dispatch.args[1][0]).to.equal(fetchBookStub);
-        done();
       })
       .catch(err => { console.log(err); throw(err); });
   });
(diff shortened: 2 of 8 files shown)


packages/opds-web-client/src/components/Book.tsx
```diff
@@ -208,7 +208,7 @@ export default class Book<P extends BookProps> extends React.Component<P, {}> {
       );
       links.push(
         <BorrowButton
-          key={this.props.book.borrowUrl}
+          key={`${this.props.book.borrowUrl}-borrow-button`}
```
EdwinGuzman: Just to fix a "key' issue. Most of the time `this.props.book.borrowUrl` is undefined.

packages/opds-web-client/src/components/__tests__/mergeRootProps-test.ts
```diff
@@ -5,8 +5,8 @@ import ActionCreator from "../../actions";
 
 // synchronous actions for simple testing
 // of createFetchCollectionAndBook
-let fetchCollectionStub = stub().returns(new Promise((resolve, reject) => { resolve({}); }));
-let fetchBookStub = stub().returns(new Promise((resolve, reject) => { resolve({}); }));
+let fetchCollectionStub = stub().returns(Promise.resolve({}));
+let fetchBookStub = stub().returns(Promise.resolve({}));
```
EdwinGuzman: Just minor code improvements that don't change functionality.

packages/opds-web-client/src/components/__tests__/mergeRootProps-test.ts
```diff
@@ -57,7 +57,7 @@ describe("createFetchCollectionAndBook", () => {
     actionBookCollectionStub.restore();
   });
 
-  it("returns fetch function that uses the provided dispatch", (done) => {
+  it("returns fetch function that uses the provided dispatch", async () => {
```
EdwinGuzman: Also updating the tests to use async/await.

packages/opds-web-client/src/components/__tests__/mergeRootProps-test.ts
```diff
@@ -121,31 +112,48 @@ describe("mergeRootProps", () => {
...
+      };
+      props = mergeRootProps(stateProps, dispatchProps, componentProps);
+
+      await props.setCollection("loans url").then(data => {
+        expect(data).to.equal(groupedCollectionData);
+        expect(fetchCollection.callCount).to.equal(0);
+        expect(clearCollection.callCount).to.equal(1);
+        expect(loadCollection.callCount).to.equal(1);
+      }).catch(err => { console.log(err); throw(err); });
+    });
```
EdwinGuzman: This is new.

aslagle:
This is a lot more complicated than I was expecting.

For 1, I think it would make more sense to put the catalog root link in the header if it exists, which I think the patron app already does.

For 2, the loans page should show a loading message if there's a borrow request that hasn't completed yet.

aslagle:
Haven't looked at this carefully, but: I think the original issue was only meant to stop a borrow request and loans request both being made during a borrow, not to prevent all repeated loans requests - we do want to update the loans when you visit the page in case a hold became available or you did something from another device. In general I think we should be doing more caching of different feeds but that's a bigger issue.

EdwinGuzman:
For 1, yes it does but this is for the breadcrumbs. In the stand-alone opds-web-client app, there would be no way to get from the holds page back to the catalog page.

EdwinGuzman:
What do you mean to prevent a borrow request and a loans request from occurring at the same time? How is that possible? 

I think the JIRA ticket is not giving a clear description of what the bug is.

aslagle:
I think the idea of that issue was that after the app made a successful borrow request, it was immediately making another request for the loans feed, even if you weren't on the loans page and didn't need to update the status of your other loans. Maybe it was already changed and it's not doing that anymore - I don't remember exactly where the code was, but looking at https://github.com/NYPL-Simplified/opds-web-client/blob/master/packages/opds-web-client/src/reducers/loans.ts#L36 it seems possible it already works.

EdwinGuzman:
If I understand that correctly, that only updates the array of books in the loans feed to include the new book. So this doesn't make a request to the loans feed. The code I added in makes allows this newly borrowed book to appear in the loans feed without making a new request, but it does have the drawback that updated information for other loans don't show up.

This seems like an issue that's not really an issue anymore. Where else can a user see the array of loaned books besides the "My Books" loan page?

aslagle:
They can’t see it anywhere else but when they look at books in the regular catalog they can see if they have them on loan. You’re probably right that it’s not an issue anymore - that code I linked may have been the fix for the original issue

EdwinGuzman:
Cool. I'm going to update the ticket as "won't fix" and write the issue may have been fixed already. Thanks, closing this.

PR #242, opened 2019-11-07
Fixing accessibility issues found by react-axe.

EdwinGuzman:
Resolves [SIMPLY-2381](https://jira.nypl.org/browse/SIMPLY-2381).

A lot of issues were caused by color contrast and to fix that I imported and slightly darkened the blue color we've been using in the Reusable Components repo. This does affect the link colors and doesn't make them as bright, but fixes accessibility issues. This also affects buttons where the link color was used as the background color:
Screen Shot 2019-11-07 at 10 27 46 AM

Another big change is that all form inputs now have a label. Maybe the most noticeable one will be the search input in the breadcrumbs:
Screen Shot 2019-11-07 at 3 21 57 PM
Screen Shot 2019-11-07 at 3 21 42 PM

Navigational elements didn't have the `role` attribute in the correct element. This was for `ul`/`ol` elements which are now wrapped in a `nav` element.

The `role="complementary"` was not needed and should be included in a top-level element, not an element that is nested. In this case that occurred for the "facets" element in the `main` element.

Also updated syntax for react refs in the applicable components.


packages/opds-web-client/src/components/BasicAuthForm.tsx
```diff
@@ -23,19 +26,21 @@ export default class BasicAuthForm extends React.Component<BasicAuthFormProps, B
             { this.state.error }
           </div>
         }
+        <label htmlFor="login-input">{this.loginLabel()}</label>
```
EdwinGuzman: Removing the placeholders in favor of labels:

Screen Shot 2019-11-07 at 3 28 15 PM

aslagle:
The guides I'm finding say it's fine to hide a redundant label for an input or use aria-label instead, e.g. https://www.w3.org/WAI/tutorials/forms/labels/#hiding-label-text and https://a11y-style-guide.com/style-guide/section-forms.html#kssref-forms-search

packages/opds-web-client/src/components/Collection.tsx
```diff
@@ -47,7 +47,7 @@ export default class Collection extends React.Component<CollectionProps, {}> {
     return (
       <div className="collection">
         { hasFacets && (
-          <div className="facet-groups" role="complementary" aria-label="filters">
+          <div className="facet-groups" aria-label="filters">
```
aslagle: If "complementary" isn't right, maybe this should have some other role to make it easier to navigate to.

EdwinGuzman:
Yes, I noticed that while working on `circulation-web` and the select in the header. Will make the search input have an aria-label, but I think the rest of the labels are fine.

packages/opds-web-client/src/components/Collection.tsx
EdwinGuzman: This makes sense as an `aside` (instead of a div) and I'll try that

EdwinGuzman: Aside doesn't work either. This is the issue: https://dequeuniversity.com/rules/axe/3.3/landmark-complementary-is-top-level

Removing the `role="complementary"` passes the aXe test. I think this should be its own ticket which can be worked on later since this is a bigger change. It'll require updating the app so that this section is no longer in the main container and be able to be an aside or have the "complementary" role. What do you think about this, rather than updating it in this PR?

packages/opds-web-client/src/components/Breadcrumbs.tsx
```diff
@@ -18,21 +18,23 @@ export default class Breadcrumbs extends React.Component<BreadcrumbsProps, {}> {
...
-                  collectionUrl={link.url}
-                  bookUrl={null}>
-                  {link.text}
-                </CatalogLink>
-            }
-          </li>
-        ) }
-      </ol>
+      <nav aria-label="breadcrumbs" role="navigation">
+        <ol className="breadcrumb">
```
adriana-alter: It seems confusing that the `ol` rather than the `li` is called "breadcrumb"; I realize that didn't come from this PR, but I think calling the nav "breadcrumbs" makes it even more confusing.  I would go `nav.breadcrubs-nav', `ol.breadcrumbs`, `li.breadcrumb`.  But this may not be worth messing with; it's not at all a big deal.

packages/opds-web-client/src/components/Search.tsx
```diff
@@ -29,6 +29,7 @@ export default class Search extends React.Component<SearchProps, {}> {
             <input
               className="form-control"
               ref="input"
+              aria-label="Enter search keyword(s)"
               type="text"
```
adriana-alter: The `(s)` seems annoying if you're using a screen reader.  "Enter search keyword or keywords"?

adriana-alter:
This is REALLY minor, but in the breadcrumbs, the links' vertical positioning looks really weird to me, since they're aligned with neither the label nor the search box.  I think dropping them to the bottom (i.e. aligning them with the search box) would look nicer.

EdwinGuzman:
@aslagle are you okay with these updates?

aslagle:
I don't have time to look at them but I trust you

PR #243, opened 2019-11-12
Typescript upgrade

kristojorg:
This PR simply upgrades the typescript version in oder to fix some issues I was experiencing when using `opds-web-client` within `circulation-patron-web`. In order to upgrade typescript, I also upgraded `tsloader`, `prop-types`, and a few type declaration packages. I then fixed all the type errors I saw from breaking changes to typescript.

A few formatting changes were made on accident due to vscode's lint setting `formatOnSave`. I have disabled that and attempted to revert most of the format changes to make the PR easier to read. 

Let me know if you have any questions or edits you'd like me to make. 

### Breaking change

I was not able to get this to work with node v8. The current node version is 13, and I'm wondering if there is anything specific holding us back from upgrading that?


packages/opds-web-client/src/components/CatalogLink.tsx
```diff
@@ -12,20 +12,20 @@ export interface CatalogLinkProps extends React.HTMLProps<{}> {
 export default class CatalogLink extends React.Component<CatalogLinkProps, {}> {
   context: NavigateContext;
 
-  static contextTypes: React.ValidationMap<NavigateContext> = {
```
EdwinGuzman: What happened to the `NavigateContext` type?

kristojorg:
Thanks for catching the `NavigateContext` change, @EdwinGuzman. I did in fact mean to ask about this before submitting a PR. I'll continue the discussion here because it is a little lengthy and I find it easier in the main conversation. When I upgraded `typescript`, I started getting the following error on `router`. 
```
Type 'Validator<object>' is not assignable to type 'Validator<Router>'.
  Type '{}' is missing the following properties from type 'Router': push, createHref, isActive
```
it seems that the `PropTypes.object.isRequired` type is marking `router` as an empty object, so typescript is telling me it is missing keys necessary to be deemed a `Router` type. I tried adding the members to the type:

```
static contextTypes: React.ValidationMap<NavigateContext> = {
    router: PropTypes.shape({
      push: PropTypes.func.isRequired,
      createHref: PropTypes.func.isRequired,
      isActive: PropTypes.func.isRequired
    }).isRequired as React.Validator<Router>,
    pathFor: PropTypes.func.isRequired
  };
```

But then got this error:
```
Type 'Validator<InferProps<{ push: Validator<(...args: any[]) => any>; createHref: Validator<(...args: any[]) => any>; isActive: Validator<(...args: any[]) => any>; }>>' is not assignable to type 'Validator<Router>'.
  Type 'InferProps<{ push: Validator<(...args: any[]) => any>; createHref: Validator<(...args: any[]) => any>; isActive: Validator<(...args: any[]) => any>; }>' is not assignable to type 'Router'.
    Property 'push' is optional in type 'InferProps<{ push: Validator<(...args: any[]) => any>; createHref: Validator<(...args: any[]) => any>; isActive: Validator<(...args: any[]) => any>; }>' but required in type 'Router'.
```
which for some reason is telling me that `push` is optional, even though it's marked at `isRequired`. 

One way to solve this is to just cast the object in place:
```
router: PropTypes.object.isRequired as React.Validator<RouterType>
```

I'm curious what you would like me to do. I am not sure that spending time typing this static property is super valuable since the type here doesn't flow through to when you're using the context value... that type is taken from the static property `context: NavigateContext;` which is declared on the class already. This is only used to tell react it needs to make the context available here.

packages/opds-web-client/src/components/CatalogLink.tsx
kristojorg: See my reply in the main thread @EdwinGuzman

EdwinGuzman:
I think not spending too much time on it is fine and casting it seems reasonable right now.

packages/opds-web-client/src/components/BookDetails.tsx
```diff
@@ -3,7 +3,7 @@ import * as moment from "moment";
 import BookCover from "./BookCover";
 import Book, { BookProps } from "./Book";
 
-export interface BookDetailsProps extends BookProps {};
```
EdwinGuzman: So I'm guessing this is now unnecessary Typescript?

packages/opds-web-client/src/components/CatalogLink.tsx
```diff
@@ -13,19 +13,19 @@ export default class CatalogLink extends React.Component<CatalogLinkProps, {}> {
...
-    router: PropTypes.object.isRequired
+    router: PropTypes.object.isRequired as React.Validator<RouterType>
   };
 
   // provides full router context expected by but not actually used by Link
   // see https://github.com/reactjs/react-router/blob/master/docs/API.md#contextrouter
   // and https://github.com/reactjs/react-router/blob/master/modules/PropTypes.js
   getChildContext() {
-    let noop = () => {};
+    let noop = () => { };
```
EdwinGuzman: Is the space necessary or  just something that wasn't reverted?

.gitignore
```diff
@@ -12,3 +12,4 @@ packages/opds-web-client/docs
 
 # Temporary files
 *~
+node_modules
```
EdwinGuzman: This should not be necessary (but doesn't hurt) because the root level `package.json` doesn't install any packages.

EdwinGuzman:
I tested this by linking it to `circulation-web` and `registry-admin` and there was only one issue in `circulation-web` which was easy to fix. So this is great and I can update Typescript in those other apps later on. I just have some minor questions.

packages/opds-web-client/src/components/BookDetails.tsx
kristojorg: I'm actually not sure. The only change I made here was to remove the semicolon because the linter complained when I tried to run the tests. Doesn't relate to my PR though, so I'm not sure how the errant semicolon got through in the first place

packages/opds-web-client/src/components/CatalogLink.tsx
kristojorg: Just something that wasn't reverted. I can remove it

EdwinGuzman:
Great, so let's see what @adriana-alter thinks and then we can merge it in.

adriana-alter:
This looks good to me!

EdwinGuzman:
@kristojorg all PRs should have an updated package version in `packages/opds-web-client/package.json` (so we can publish to npm) and an updated `changelog` based on its version and its update. Since I have a PR which I will merge soon #242 , I think it's fine if these two PRs are bundled into one new version bump (so no need to update the version and changelog in this PR). I can merge this first and then fix any merge conflicts in my PR. Sounds good?

kristojorg:
@EdwinGuzman that sounds good, I remember you saying that... just was waiting I suppose until we got merge approval to make those changes. If you want to bundle them that works for me too!

.gitignore
kristojorg: Ah yes, it's actually part of my `prettier` PR where this becomes necessary

EdwinGuzman:
Great. I'll bundle them together. Thanks!

EdwinGuzman:
FYI, @kristojorg and @adriana-alter , the next version of the `opds-web-client` I'm going to publish will go from `0.2.10` to `0.3.0`. This is a bigger change than usual along with some accessibility updates and I think a minor version bump makes sense.

kristojorg:
Thanks @EdwinGuzman !

PR #244, opened 2019-11-12
Prettier

kristojorg:
This PR is a suggestion for you all to take a look at. It adds https://prettier.io/ to format all code in the repository according to a central source of format rules. I find it to be surprisingly helpful (especially when working in teams) for a few reasons:

1. All code will get formatted pre-commit with a [husky](https://github.com/typicode/husky) pre-commit hook. This means that from now on, PRs will never be cluttered up with formatting changes. If a file works the same, it will look the same in the PR (with the exception of comments obviously). Very helpful

2. You don't ever have to worry about formatting your code nicely again. It's a pain to have to add line breaks to function parameters when the function gets too long. Or to have to indent pasted code properly, etc. Instead, I set my editor to run prettier on save,  so I just paste a blob of code and hit save, or add a few parameters to a function and if it's too long, prettier will organize it for you.

3. Finally, code is much easier and much faster to read when it's all formatted the same

Read more about why prettier [here](https://prettier.io/docs/en/why-prettier.html)

So this adds the pre-commit hook and runs `prettier` on all `.ts`, `.tsx`, `.json`, and `.scss` files in the repository. I configured it not to add spaces around brackets because I've heard @EdwinGuzman prefers that. So `const {req, res} = params` not `const { req, res } = params`. We can change other things as well, such as configuring the width that prettier tries to keep lines below. 

Let me know if you have questions and what you think. I branched this off my `typescript-upgrade` branch so it's a PR back to that branch.


adriana-alter:
This looks very nice to me!  How difficult is it to set up?

kristojorg:
@adriana-alter great! It's set up as a pre-commit hook on this PR. Was super easy to make that happen. From there, you can add an extension to your editor to do format on save, which is really helpful also

EdwinGuzman:
It looks good so far but, sorry, I meant the spaces in braces looked weird for inline jsx but not for anything else. So import statements and variable declarations `const { req, res } = params` are fine. But, this is my preference and I'd like to hear what @adriana-alter thinks.

adriana-alter:
I like the spaces too.

packages/opds-web-client/src/AuthPlugin.ts
```diff
@@ -1,14 +1,26 @@
...
+    credentials?: AuthCredentials;
+    error?: string;
+  } | void;
+  formComponent: new (props: AuthFormProps<AuthMethod>) => React.Component<
+    AuthFormProps<AuthMethod>,
+    {}
+  >;
+  buttonComponent: new (props: AuthButtonProps<AuthMethod>) => React.Component<
+    AuthButtonProps<AuthMethod>,
+    {}
```
EdwinGuzman: Are these being broken up because it's a long line or because each param should be on its own line?

packages/opds-web-client/src/components/__tests__/DownloadButton-test.tsx
```diff
@@ -101,18 +107,25 @@ describe("DownloadButton", () => {
     expect(fulfill.args[0][0]).to.equal("download url");
   });
 
-  it("opens indirect fulfillment link in new tab", (done) => {
+  it("opens indirect fulfillment link in new tab", done => {
```
EdwinGuzman: I remember that this was something that the linter complained about. Does prettier override those complaints?

EdwinGuzman:
This is great. Can you just fix the merge conflicts and the tests?

packages/opds-web-client/src/AuthPlugin.ts
kristojorg: Because it's a long line. They try to keep lines below their `printWidth` setting, which is currently 80. We can set it to longer if you like. There are exceptions where they won't break the line up, like a string literal for example.

packages/opds-web-client/src/components/__tests__/DownloadButton-test.tsx
kristojorg: Yea the idea is that prettier will allow us to disable all lint rules that deal with formatting. So you'll never get complained at for that again, because it's formatted automatically. 

In terms of this specific formatting option, we can set it to either have the parens or not have them on single argument arrow functions, totally up to you which you prefer. No parens was just the default. It of course will always have parens when there are multiple arguments

kristojorg:
Okay yea I'll fix the tests and hopefully will be able to resolve the merge conflicts. The one tricky thing with setting this up is that it will create merge conflicts between master and any other working branch, since it effectively alters every file. There are two ways we can resolve this:

1. Add prettier to all working branches independently. This would be simple enough, and would probably actually reduce the number of merge conflicts when ultimately merging into master since all formatting will be the same. The process would be to (1) get rid of inactive branches, (2) I'll add and run prettier on every remaining branch. 
2. The other option is to perform a series of rebases whenever trying to merge unformatted branches into master. It's a bit complex, but also doable. My preference would be for option 1. 

For now, I'm going to redo this PR onto master with the changes you asked for

kristojorg:
I'm closing this to redo these changes on top of master.

PR #245, opened 2019-11-13
Logic components

kristojorg:
This PR makes a few 100% backwards-compatible changes that will make it easier for `circulation-patron-web` to consume `opds-web-client`. There are three things it accomplishes:

1. Pass the store down the tree in context. This did not appear to be done previously, so I'm actually not sure how `react-redux`'s `connect()` function was able to work, since it typically looks for the store in context. If I'm missing something, and it is already being passed down, let me know! 
2. Separate out context providing components from visual components. I created a wrapper component to pass `pathFor` down the tree without rendering anything. This makes it usable by `circulation-patron-web`, and is a better separation of concerns. 
3. Create a `RouterContext`. `circulation-patron-web` is upgrading its version of `react-router`, and the new version doesn't pass `router` down the tree in context, so I'm going to do that myself. I've created the `RouterContext` here so that `opds-web` could also use it if you wish. It would make sense ultimately for `opds-web-client` to own this component, but if you'd rather I move it out to `circulation-patron-web` for now, I can (it's not currently being used here). Alternatively, if you'd like me to update `opds-web-client` to use this component to pass the router down the tree, that is also possible. The benefit would be that the `router` would now be accessible via the new context API as well as the old, and you could upgrade `react-router` with few changes.

Both context providers pass their context down via both the [old context api](https://reactjs.org/docs/legacy-context.html) and the [new one](https://reactjs.org/docs/context.html).

I also add the top-level `node_modules` to `.gitignore`. Correct me if that wasn't meant to be there, but I was getting errors having so much to commit after installing.

The PR is to `typescript-upgrade` since I branched off from there.


adriana-alter:
This looks very nice; even if passing the store down isn't strictly necessary, we might as well go the more Redux-y route!  I'm okay with this `RouterContext` implementation for now.  Can you update the tests?

EdwinGuzman:
It's looking good and glad the separation of concerns is being done. It looks like those components only do one thing, which is setting up and passing down context. I need to brush up on the docs for the way the new context API works to fully comprehend this.

If I understand this correctly, `PathForProvider` wraps `OPDSStore` and both set up their own context for separate variables. Does the order matter?

Leaving `RouterContext` here is fine and I wouldn't mind integrating that context here and upgrading RR in a separate PR. Since you don't need to do that for `circulation-patron-web`, you don't need to worry too much about it.

EdwinGuzman:
This update works fine in `registry_admin` but I get these errors in `circulation-web`:
Screen Shot 2019-11-14 at 12 30 02 PM

It's probably not difficult to update `circulation-web` to make it work, but this may be an issue that doesn't make this update backwards-compatible. I would like to fix those errors first before merging this in (not that we will _have_ to use this updated version right away, but I'd feel better to have apps up-to-date).

kristojorg:
@EdwinGuzman ok I'll take a look at `circulation-web` to see why that's happening. The order or `PathForProvider` and `OPDSStore` only matters because `OPDSStore` depends on the context that `PathForProvider` is passing down

I'm working on the tests now and then will take a look at `circulation-web`

packages/opds-web-client/src/components/OPDSCatalog.tsx
```diff
@@ -18,28 +16,30 @@ export interface OPDSCatalogProps {
...
+/**
+ * Simple wrapper to allow store to become available in context
+ * before we try to access it.
+ */
+const RootWrapper: React.FunctionComponent<OPDSCatalogProps> = (props) => {
+  const {store} = useStore();
+  return (
+    <Root store={store} {...props} />
+  );
+};
```
EdwinGuzman: Is there a reason why the `RootWrapper` exists. It seems possible to render `Root` in the `OPDSCatalog` function and call `useStore` in there, right? Just wondering why the need for this extra `RootWrapper` function.

packages/opds-web-client/src/components/context/StoreContext.tsx
```diff
@@ -0,0 +1,38 @@
...
+  constructor(props, context) {
+    super(props);
+    this.store = buildStore(this.props.initialState || undefined, this.props.authPlugins || [BasicAuthPlugin], context.pathFor);
+  }
+
+  render() {
+    return (
+      <Provider store={this.store}>
+        {this.props.children}
+      </Provider>
```
EdwinGuzman: Why doesn't this use a similar pattern like `RouterContext.tsx` where there is a `createContext` variable to use in the render function? Is it because the `store` attribute is explicitly available in this provider?

EdwinGuzman:
Okay, this makes a lot more sense to me now after reading up on the old/new API. Can you just write tests for the changes in this PR, please?

EdwinGuzman:
To be clear, when you mentioned

> Pass the store down the tree in context. This did not appear to be done previously, so I'm actually not sure how react-redux's connect() function was able to work, since it typically looks for the store in context. If I'm missing something, and it is already being passed down, let me know!

did you mean that there was previously no `<Provider store={store}>...` wrapper component for the store to appear in the app's context?

packages/opds-web-client/src/components/OPDSCatalog.tsx
kristojorg: You might be right actually. I thought you couldn't access the context unless it was separated by an encapsulated component like this, but maybe that was with the old context API or I'm misremembering. Just tried it in a [codesandbox](https://codesandbox.io/s/adoring-pasteur-hu5m5?fontsize=14&hidenavigation=1&theme=dark) without that and it seems to work. I'll remove this

packages/opds-web-client/src/components/context/StoreContext.tsx
kristojorg: Yes exactly, it is because the `context` variable is already created by `redux`, we are just importing it, rendering the provider, passing in a value to the provider, and also passing the context down via the legacy API as well. Basically, redux already created it for us

kristojorg:
>did you mean that there was previously no <Provider store={store}>... wrapper component for the store to appear in the app's context?

Yep exactly. I'm not actually sure how the store is appearing in the app's context (or even if it is). Redux's `connect()` function relies on accessing the store from context, so I'm not sure how it's currently working because I can't find anywhere the store is passed in to context (previous to this PR)

EdwinGuzman:
Why was this closed? I think it works well, unless if you're going to do this differently?

kristojorg:
@EdwinGuzman I have no idea actually. That must have been an inadvertent click on my part. I'm currently getting the tests fixed and then going to solve the issue you were seeing and then this will be ready. I don't think I can reopen it though, can you do that for me?

PR #246, opened 2019-11-14
Set up prettier on master

kristojorg:
This is a redo of my previous "Prettier" PR, basing off of master. If we accept it, I'll then go through and add prettier in a similar way to every other active branch to avoid extensive merge conflicts.

This adds a few dev dependencies to the root `opds-web-client/package.json`. It does this in order to run prettier on all code in the repo at every commit, thus ensuring that once code reaches github, it is all formatted properly. `husky` is a package to run the `pre-commit` hook. `lint-staged` is a package to run a command on staged files. This is the setup recommended by the folks at prettier. I add `node_modules` to `.gitignore` because we now are installing packages at the root level. 

I also removed the bracket spacing rule I had added, so brackets will have spacing now - we are now using the prettier defaults without changing anything.


EdwinGuzman:
Overall I like it but am concern about a few things. The 80 character line limit I like and try to follow but not sure if strictly following it will be good. A few things it breaks make lines slightly harder to read. Maybe we can try 100 instead of 80.

I'm not sure why it breaks things up like `response.text().then(text => {` into 
```
response
    .text()
    .then(text => {
```

but it converts
```
fetchMock
    .mock("test-url", 200)
    .mock("http://example.com", 200);
```
into `fetchMock.mock("test-url", 200).mock("http://example.com", 200);`

I guess because one is a promise. I'm okay with these changes if we all agree to it, and especially @adriana-alter since it should also be implemented in the other apps.

adriana-alter:
I don't love the difference between the `response.text()` and the `fetchMock` either; it's pretty weird--and too arbitrary--for promises to have different formatting rules.   Are there any settings that we can mess with to make that consistent? 

As far as the 80-character limit: is there any way to set Prettier to flag specific suggested changes for review, rather than just going ahead and changing them, i.e. is there a convenient way for us to confirm that we want all those lines to be broken up, rather than having to go through afterwards and revert things manually?  If not, I like the idea of making it 100 characters.

Otherwise, I'm fine with this!

EdwinGuzman:
I think the settings would be global and we may not be able to pick and choose where it is enabled (or revert), but I could be wrong. The 80-character limit is very common in projects and a general dev setting, but I think it'll be worth going with a 100-character limit for us.

kristojorg:
The failing tests are coming from `sass-lint` errors due to conflicts in how prettier vs sass-lint prefer the `.scss` files to be formatted. We can either (a) not format .scss files with prettier, or we can (b) upgrade sass-lint for [style-lint](https://stylelint.io/). Sass-lint is no longer being maintained and so doesn't have a way to integrate with prettier, while style-lint is actively maintained and provides a plugin to integrate with prettier rules. I'm not really sure how difficult the upgrade would be, but could give it a try if you'd like.

In terms of the particular formatting points you bring up... I just saw this on the prettier website when poking around the options:

> For readability we recommend against using more than 80 characters:
> In code styleguides, maximum line length rules are often set to 100 or 120. However, when humans write code, they don't strive to reach the maximum number of columns on every line. Developers often use whitespace to break up long lines for readability. In practice, the average line length often ends up well below the maximum.
> Prettier, on the other hand, strives to fit the most code into every line. With the print width set to 120, prettier may produce overly compact, or otherwise undesirable code.
> See the print width rationale for more information.

adriana-alter:
Okay, the chaining thing makes sense; I'm on board with that.

I still don't love the 80-character thing, but it's certainly not a dealbreaker.

Is there a way to retroactively override changes, or flag things that shouldn't be reformatted--i.e. if Prettier splits up a line that shouldn't be split up and I fix it, will it just get split up again every time I hit save?  I realize the point is so that developers won't have to think about formatting, but I would imagine there's still some way to make exceptions if necessary...no?

Sounds like we should upgrade to style-lint anyway.  Unless @EdwinGuzman knows of a specific reason we haven't done this already, branch off of this PR and give it a try!

EdwinGuzman:
Yeah, I'd be okay with going with an 80-character limit. The chain formatting makes sense but I think, @adriana-alter , that when you hit save or commit, it will be split every time.

@kristojorg , for now, can you not format scss with prettier and we can make a ticket to update that soon. It may not be worth your or Amigo's time to do that (unless if it's a very easy change).

kristojorg:
There are certain cases where prettier will result in slightly uglier code, but I think these are edge cases, and the many advantages far outweigh those few times for me... especially in a react codebase where formatting jsx can be specifically time consuming and annoying. 

[Here is a playground](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEBLAtgBwgJxgAmHzgA9M4wCBffAMxwnXwB1wALAQ1VYG5moM2PIXwBnGAFcARvhr1GLEKNRRovfpCji6cGGDYBZCGADW+ALz4ccAI4TU1gBStau-QFp0xk6wCUfKH5BXAIAEQ4YDgAxNzY4HDoGJlYAOhSAenDImL04nHUoTW0AYQgIE1Q4CytbeydWACtRd0hyyr8A-gATOFEwHFQpOGcQLOjY+NYAGnxHXwsAPkJ+fHwigg4ujkx4BMtHLciZiRwAG3nzJdZN7d2C1Z6+gaGR11y56dmLpeAV1fwhrRcHAAKIcfSOOaLZZQf7-N76IymFJeUwjeDidwnU6fABMAAZ8b4Ud4RmwYDBMEh0ulSBwsKc4ClIOg8YT-H9VlQOYFYf8OLRdmCIVDLjC4asEYZvClrOJgXMAnDuZ0+atUDARhJRL0dO9fJ9RT9Of9GQQIDtUNBRNVfmqJehdGwIF0kIoAAoAeQAygAVaYmuGHDhuogYmBu1jh1iyKaB-79OA9WCoDinUSRpT0uDuXCoADmKlY8aoSolZr1+ni1SgcAA7vgxjkqzhHEQbjtqyr41L4ikpejejAsWdPhaYFatDyJfgK1KAII4fM2yxSpEmZlp9OK-jx0jkSiONcysBbpNzYkwCApWppxwwHASODTiX7iiahdL0QAbXxAF1fz-S9r1vU4RnScMRxxEAXzhN9D0-ZdAO-ABGICUivFIejgTAbzsO87RnVY0hScdJ1EON7ThOJNniDMRFYAANdwACVaiHJN3AAdQ1NhWEzRiDAAGQACQpTA2LsIcYyoEtYNkHkTQ1EYdSgLobUTZMJzTG0pAAT3wHpaA4CRThgQ1vnFctdHwMjrVteNVkdGBnVdD0fX9EBKKIwyIhDERw0zaMQFkEsyzhOcJj2fBawbJsorbfAO12BTwvhKL+1iQdMWxMdLWteTVki3JF2Xapj2RU9Tm3RSqOIMh3yPWJ103arz18YC8IkO8HyfQr6oPD9YlKn9-0AzrQPAyDcpgtLVngoaSq-ZC0M67DcMmwiiM0hBtPTTNRGzXMBkLQIvMc-AaJ6HB6K2nymNY9jxE4niXP4xRBNE8TJKfcRizqrlvKIki7K0OS0u7Xk4WU1htV1TAGBIAzsXwVBaHwBGIAAN1QHougs6E7oTU4OFEG11yiXB0DGKyZ2DN0OCgPTVR89ZH0oXAoSJmcXNQUQsL821Swu2SoaI7ZyDUxwTDgPSZixtN5m5iVef54NvxlvS-2qBXTjm-5RYu-NdGl2WlYu1ZrEkHBYSIXWnzdVWBciDXZe14WAdCsWZwl049K5i2amtqB9a5EtdzqucqbGb1JBkSxxGkRw6xULoIDrGZWEpnBqb8vxWvTKIOBl5xPaNGL63wCno78j46tqmcK0xpGAFUzmqVhyUpalaRIelMEZZlGHuayCF7aLYsbPzmzyRLm70tvThmZKu368fMveKMhygjp42K-QRvK5qTy3UQd29-4Fqa3IWqqxkugvDCQPwsDeuffWr8Q0aAP-CaX8cAABgAEmAPPReVB0gAP6p-Yay0xpoRRE6F0f9upgVYF6P0u86owKWkheBf4UhSBdHpFIxtNSw1HDBFI9tnxPy6neLeOVKENwlECHOMc46yiHAqeSkMlLkJAAjHCHA5RJQkC5S6cBaI3QJmKZW+88g1krvFXI8REorwSJDRuNkdoph0raDGDAcbXSCtJLyaxrBaVTPtRQui9qiBkvrdeZD5ziLYMUSxu1rEri+NCOx3inEZQHIw4cM1+oKMPquY+lVT7nxnDgg+cCAIIKunRb8rBXEuTzAALwiJOVg6FMKTVYP4nSWDlQvj4XyR4-RBjDAyW4-AHikxeLKeY8uRMYYgDTr0KAAByAgOoNiNNKemVG6MXLxCqCIqoqhayyONJHGy48lFxWngldsWxOyaLXhlIZmT3GeL0duCQak4C0BUEmaBDVDylDaEyMh18WwpBMi5AA0mbNacAcL0LAqcoylyui8JYeqARQybSvLYBYlpxyHHtMsvI5ZUVVlT2yBspKWyUpaNHtCqx+jLBEExsY+Ipi-rmNGfREpRz7GOJ7Hs3QBzml4u3BS65g1HB3IqA8k269IUfL0h1Oh61fmlx8gAKW9J6AAcikcQAwoCFloP7Vl8ZgUs3+F0shEKRnUu8Qs2mpokWqInso9Zxr1GYtXvrCsFKDFEtxiSxQwUZgUszBS2ldVOWVFlTyjKfLZYzAldK2VD4VCKuVbqnSHUP43MWs8lxbimWtJql8n5k1WUQxBajARYBGQiO1RI918LCZ7yNS2FFKiWwWtuFa0tBBbUEsMdjB1OBSXmXJZGmxVKYU0pCh7GcXqmRDKeXkF5bj+WBslTKuVYa0YRp7d46NdLjXMjzTgRlnaz5ssaoO0hvqV3+oFSgu8-zzmArVX8KpDxei1JeC4WInp3ShG9GMD4xa5H8JGNYBo74IVqSDicLQSVYTxAYAkNG+BJkYpreB0Q-T63QAufmE4SZPikz0oUXxH6llj2RZYSelbZ5ZtWDSfAUqIAVyTJBijqIzAcErFCuU2AtBVGUIUKoGoxDOlMl0GKEACDWHBFClyEQMYQBUDAFI8YOB1i4LhldUpH3PtfSEneVDTzvFAyWz2V9QMhoiNqY9YEoCmXOKHAajU9NMetLQop-9A6sHI9B7Zl1SYAm+bCTQSGUO8ZUKi8YK7-o+X6vNWNjg9PYiM9lUJzD9bYq5FmrpLkGB1htKB3A4yaiiGYzqVGcGBn4AJPiNDogMNgCw4smcFUNy0ei2pmYRBxAGfogAFnxChGYRCuh6UzKcyFOTUN9uI7OMtij8OmrReazZMHUrSdk5x9eimn0vtrqpsJm5NM4D2JVoiumtv6ckKIKLbWUIhYs4eKzvQcu2efqgrUUB+sDGyYNs7e2cApEi3Q4pIBpqxfBhHKr0SNxyivE4C9F8ktsBS3+4gW2MsQag9ZljeW+MECU96ErZWKsGslEDkkaI1uUIa2ISIh23RFc68QzMqg0fLZksNhR1ZxtrMm1W6bLn4v8nm-J55S3lOrZ+9vdbGmIRaewz5N7B3DNff-iZ6qr2wuXeyzZqLWcuD32oxjfNVR0e+UiOU3biv9ufbs3dwn0F+pXtYXjkHPDM3qtBeiKH6cYfpYSNAJKAJNhrC3JjzDHTl2IhlLV83nxGuk+1G6AArISSn3XqcWjgPTh3hqedjYrizgLbPnNYv6jJuTDG+x85W5EOrwuIii629piXRv3tI51KrkAqhyAG4qSn3HN8ZS27B-by9PJIZeRAKDUQyBQAiJS+6GZWhkAgA4FjcTXRB9SBwOCGWMBvSYHBGG5Ab8pggHJOgU4XE2Aaiu+COA3oEDKAnDjGAekZ9gDJoPlQOo8DuhX-mdAHBkDGXTHAPfTQJAAAQivqYLoN6NmEJJcj-jpP-iAIAd6GGoyAAIoSD8ZwAwF-576b43TxAz5SAcBDCnCD4IwSY8RdAuTIAoSx6CIMA6hcQr6YAz5CKv5YwYF752DoHv4Wgj4oCkzuC1gtKL577WB2AOBwDv4cCf7f5IC-46h746joCoA76PhwFsb5goFoHwCYHyE-aEHkGUFIC4h74Pjq5hqlDoBf4z5zLsEgBwy+iEG8FyFwFYxPgACSZysA3otSOw84akscekjIOhcAVAVAQAA) where you can try various line length settings on one of our files with a bunch of chained calls.

I'll go ahead and remove the .scss formatting from this for now. 

In terms of how to override, you can add a `//prettier-ignore` comment I believe above the line (similar to a linter) to ignore specific edge cases.

kristojorg:
Ok I rebased onto master to resolve conflicts by accepting all changes from master into this branch, then simply rerunning prettier on all those files, then force pushing my local branch back to origin. Should be good to go now! 

Let me know which branches are currently active and I'll get prettier added in to them as well so that we can avoid massive conflicts down the road.

kristojorg:
I messed up this rebase somewhere in `package-lock.json` and it's going to take more time to untangle than it will to just reapply prettier on a fresh copy of master, so that's what I did here : https://github.com/NYPL-Simplified/opds-web-client/pull/247

PR #248, opened 2019-11-22
Logic components

kristojorg:
Ok this is a redo of my previous logic-components PR here: https://github.com/NYPL-Simplified/opds-web-client/pull/245

Same methodology, just now on top of the `master` branch. Tests are added/fixed, but we still need to make sure this works well in `circulation-web` and other consumers of `opds-web-client`, which I'm leaving up to @EdwinGuzman 

We should get the `prettier` PR merged before this one. I ran prettier in here as well to avoid a bunch of conflicts, but that means that right now the "changes" are somewhat incorrect. It'll sort out once prettier is merged into master. 

@aslagle this PR is essentially beginning to separate the app into "container" and "presentation" components, like you mentioned in JIRA here https://jira.nypl.org/browse/SIMPLY-627 . (I had forgotten the term "container" components, and therefore used "logic" instead. Continuing with this will help me a bunch in `circulation-patron-web`, but this is a good start.


EdwinGuzman:
I have a PR in for `circulation-web`: https://github.com/NYPL-Simplified/circulation-web/pull/282
Just fix the conflicts (along with updating and merging in the prettier changes) and we can move from there.

packages/opds-web-client/src/components/__tests__/OPDSCatalog-test.tsx
```diff
@@ -4,73 +4,52 @@ import { stub } from "sinon";
...
+    });
+    /**
+     * This is painfully fragile, but must be done to make
+     * enzyme render the child function beneath ReactReduxContext.Consumer
+     */
+    const root = wrapper
+      .dive()
+      .dive()
+      .dive()
+      .dive();
```
EdwinGuzman: Is this because it needs to render `Root` which is under `ReactReduxContext` which is under `OPDSStore`?  That's three levels or am I counting the dives incorrectly?

EdwinGuzman:
It'll be great to get the prettier things in first to make the changes made only in this PR more readable.

packages/opds-web-client/src/components/__tests__/OPDSCatalog-test.tsx
kristojorg: Yea the dives are difficult to count, but essentially enzyme doesn't have great support for child functions and we need it to render that in order to test it properly. We alternatively could `mount()` the whole thing, but that would render all the way down the tree which would also not be a great way to test, so I thought this was the best bad option

EdwinGuzman:
Now that `master` has been updated, this PR can be updated too.

kristojorg:
I updated the PR with master so it should now show only the files I've actually changed ✨

packages/opds-web-client/CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 ## Changelog
 
+### v0.3.2
+
+- Pass the redux store down the tree via context (this was not done previously)
+- Extracted PathForContext from the Root, which provides context via both legacy and new context APIs
+- Created a RouterContext component to be used by `circulation-patron-web` to provide router context.
```
EdwinGuzman: I think "this was not done previously" can be removed or rephrase that _now_ the redux store is passed down the tree. Also, it would be good to generalize and say that any application can use `RouterContext` and an example is `circulation-patron-web`, since this component should be able to be used by more applications.

packages/opds-web-client/src/components/OPDSCatalog.tsx
```diff
@@ -18,30 +17,26 @@ export interface OPDSCatalogProps {
...
 
-/** The main application component. */
-export default class OPDSCatalog extends React.Component<OPDSCatalogProps, {}> {
-  store: Redux.Store<State>;
-  context: NavigateContext;
+/**
+ * The main application component.
+ *  - Renders root and passes props along with store to root
+ *  - Creates the redux store using OPDSStore
+ *  - consumes and passes it down
```
EdwinGuzman: Small typo, "consumes" should be capitalized to be consistent with the rest of the list.

packages/opds-web-client/src/components/context/__tests__/StoreContextTest.tsx
```diff
@@ -0,0 +1,90 @@
+import * as React from "react";
```
EdwinGuzman: The convention we are using for tests is nameOfComponent-text.tsx instead of nameOfComponentTest.tsx

packages/opds-web-client/src/components/context/RouterContext.tsx
```diff
@@ -0,0 +1,41 @@
...
+> {
+  static childContextTypes: React.ValidationMap<{}> = {
+    router: PropTypes.object.isRequired
+  };
+
+  getChildContext() {
+    return {
+      router: this.props.router
+    };
+  }
```
EdwinGuzman: Just to be clear, this is the old context API, correct?

packages/opds-web-client/src/components/context/__tests__/PathForContextTest.tsx
```diff
@@ -0,0 +1,67 @@
+import * as React from "react";
+import { expect } from "chai";
+import { stub } from "sinon";
+
+import * as PropTypes from "prop-types";
+import { shallow, mount } from "enzyme";
+
+import PathForProvider, { PathForContext } from "../PathForContext";
+import { PathFor } from "../../../interfaces";
+import { isNull } from "util";
```
EdwinGuzman: This isn't being used and `util` isn't an installed package.

packages/opds-web-client/src/components/context/__tests__/PathForContextTest.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+    const hasAccessProof = "does have access to context";
+
+    class Child extends React.Component {
+      static contextTypes = {
+        pathFor: PropTypes.func.isRequired
+      };
+      render() {
+        // here is the meat of the test
+        const hasContext = this.context?.pathFor === pathFor;
```
EdwinGuzman: What is the `?` after `this.context` for? I haven't seen it be used this way before.

packages/opds-web-client/src/components/context/__tests__/StoreContextTest.tsx
```diff
@@ -0,0 +1,90 @@
...
+    expect(wrapper.context().store).to.have.property("dispatch");
+  });
+
+  it("creates a new store if not given a state", () => {
+    let wrapper = mount(<Child />, { wrappingComponent: CombinedContext });
+
+    // now we need to see if child has access to context via new api
+    expect(wrapper.context().store).to.be.ok;
+    expect(wrapper.context().store).to.have.property("dispatch");
+  });
```
EdwinGuzman: This test is the same as the previous test.

packages/opds-web-client/src/components/context/__tests__/StoreContextTest.tsx
```diff
@@ -0,0 +1,90 @@
...
+import StoreContextProvider from "../StoreContext";
+import PathForProvider from "../PathForContext";
+import { PathFor } from "../../../interfaces";
+import { isNull } from "util";
+// import { mockRouterContext } from "../../../routing";
+
+/**
+ *  Testing the StoreContext component. Based on old OPDSCatalog Tests
+ *    - if not given state, it will build state on its own
+ *      (seems like this should be a test of buildState maybe?)
```
EdwinGuzman: I think you meant `buildStore`? I don't think there are any tests for it right now.

EdwinGuzman:
I'm going through this and learning the new Context API but it's taking me a while for it to click. 

When running the app, I get this error in the console:
Screen Shot 2019-12-02 at 4 49 57 PM

To reproduce, click on the "My books" tab in the header:
Screen Shot 2019-12-02 at 4 50 02 PM

packages/opds-web-client/src/components/context/RouterContext.tsx
kristojorg: Yep exactly

packages/opds-web-client/src/components/context/__tests__/PathForContextTest.tsx
kristojorg: It's part of typescript 3.7 called optional chaining. Basically it makes it safe to access deeply nested properties without having to test for existence at every level. So if `this.context` is undefined it won't spit out a type error and will instead evaluate to undefined. See more here https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining

packages/opds-web-client/src/components/context/__tests__/StoreContextTest.tsx
kristojorg: That's true. I actually did that intentionally because it is actually two different things that need to be tested and one test might change from the other over time. I can make them one test with a longer title if you prefer though

packages/opds-web-client/src/components/context/__tests__/StoreContextTest.tsx
kristojorg: I did mean that. And sounds good, this test suite will at least ensure a store gets built when one isn't provided.

kristojorg:
Ok yes I see this now as well when clicking "my books". It looks like `pathFor` is denoted as an optional property on `buildStore` even though it is required when an authPlugic is passed in. My error here was that in the new context api, the context is placed directly on `this.context`, instead of it being nested inside. So accessing `context.pathFor` will throw an error because `this.context === pathFor`. 

I'm not sure why typescript didn't catch this error... Do we have some setting on to make it not strict with undefined values?

In any case, updated now with all suggested changes.

packages/opds-web-client/src/components/context/__tests__/PathForContextTest.tsx
EdwinGuzman: That is amazing and really great. Thanks! That's something @adriana-alter and I can use now.

packages/opds-web-client/src/components/context/__tests__/StoreContextTest.tsx
EdwinGuzman: I see, or maybe instead of one test add a comment to explain why they are the same.

packages/opds-web-client/src/components/context/__tests__/PathForContextTest.tsx
kristojorg: It has changed my life already 🙏 !

PR #249, opened 2020-01-27
Type fixes and logic extraction

kristojorg:
Hi all,

This PR provides some type fixes that I needed in order to work with the repo properly in circulation-patron-web. It does two other things as well:

1. I created a hook called `usePathFor` which is a simple wrapper around `React.useContext(PathForContext)`, providing additional null checks that are useful so that the context consumer doesn't have to check for undefined every time they want to use the context. 
2. I extracted a bit of the logic in `DownloadButton` so that I could import and reuse it in circulation-patron-web. 

Let me know if you have any questions

### DIFF
--- a/packages/opds-web-client/src/components/DownloadButton.tsx
+++ b/packages/opds-web-client/src/components/DownloadButton.tsx
@@ -92,15 +92,6 @@ export default class DownloadButton extends React.Component<
     );
   }
 
-  generateFilename(str: string): string {
-    return (
-      str
-        .toLowerCase()
-        .replace(/[^a-z0-9]+/g, "-")
-        .replace(/(^-|-$)/g, "") + this.fileExtension()
-    );
-  }
-
   mimeType() {
     return this.props.mimeType === "vnd.adobe/adept+xml"
       ? "application/vnd.adobe.adept+xml"
@@ -108,6 +99,9 @@ export default class DownloadButton extends React.Component<
   }
 
   fileExtension() {
+    // this ?? syntax is similar to x || y, except that it will only
+    // fall back if the predicate is undefined or null, not if it
+    // is falsy (false, 0, etc).
     return typeMap[this.mimeType()]?.extension ?? "";
   }
 
--- a/packages/opds-web-client/src/components/DownloadButton.tsx
+++ b/packages/opds-web-client/src/components/DownloadButton.tsx
@@ -1,5 +1,6 @@
 import * as React from "react";
 import download from "./download";
+import { typeMap, generateFilename } from "../utils/file";
 
 export interface DownloadButtonProps extends React.HTMLProps<{}> {
   url: string;
@@ -75,7 +76,7 @@ export default class DownloadButton extends React.Component<
       return this.props.fulfill(this.props.url).then(blob => {
         download(
           blob,
-          this.generateFilename(this.props.title),
+          generateFilename(this.props.title, this.fileExtension()),
           // TODO: use mimeType variable once we fix the link type in our OPDS entries
           this.mimeType()
         );
@@ -107,14 +108,7 @@ export default class DownloadButton extends React.Component<
   }
 
   fileExtension() {
-    return (
-      {
-        "application/epub+zip": ".epub",
-        "application/pdf": ".pdf",
-        "application/vnd.adobe.adept+xml": ".acsm",
-        "application/x-mobipocket-ebook": ".mobi"
-      }[this.mimeType()] || ""
-    );
+    return typeMap[this.mimeType()]?.extension ?? "";
   }
 
   downloadLabel() {
@@ -124,9 +118,8 @@ export default class DownloadButton extends React.Component<
     ) {
       return "Read Online";
     }
-    let type = this.fileExtension()
-      .replace(".", "")
-      .toUpperCase();
+    let type = typeMap[this.mimeType()]?.name;
+
     return "Download" + (type ? " " + type : "");
   }
 }
--- a/packages/opds-web-client/src/components/__tests__/DownloadButton-test.tsx
+++ b/packages/opds-web-client/src/components/__tests__/DownloadButton-test.tsx
@@ -43,7 +43,7 @@ describe("DownloadButton", () => {
   it("shows button", () => {
     let button = wrapper.find("button");
     expect(button.props().style).to.deep.equal(style);
-    expect(button.text()).to.equal("Download ePub");
+    expect(button.text()).to.equal("Download EPUB");
   });
 
   it("shows plain link if specified", () => {
@@ -51,7 +51,7 @@ describe("DownloadButton", () => {
     let link = wrapper.find("a");
     expect(link.props().style).to.deep.equal(style);
     expect(link.props().href).to.equal("download url");
-    expect(link.text()).to.equal("Download ePub");
+    expect(link.text()).to.equal("Download EPUB");
   });
 
   it("fulfills when clicked", () => {
--- a/packages/opds-web-client/src/components/__tests__/DownloadButton-test.tsx
+++ b/packages/opds-web-client/src/components/__tests__/DownloadButton-test.tsx
@@ -6,6 +6,7 @@ const downloadMock = require("../../__mocks__/downloadjs");
 
 import * as React from "react";
 import { shallow } from "enzyme";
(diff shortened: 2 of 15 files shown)


adriana-alter:
Why are so many of the properties optional now?

kristojorg:
In circulation-patron-web, as well as in OPDS-web-client, typescripts `strictNullChecks` is not turned on, which means that even when a property is required in a function call, typescript is only checking that something is passed in, not that that thing is defined. 

In practice, many of these functions without optional properties were being used with undefined or null being passed in. In circulation-patron-web, I decided to turn on `strictNullChecks`, which meant I was not able to use them in the same way without making the properties optional. Turning this setting on has the significant benefit of guaranteeing that you will never get an undefined property access type error in runtime. 

I would suggest it is turned on in opds-web-client as well, but don't have time to fix the resulting errors. 

Long story short, these properties that are now optional were previously effectively optional because typescript wasn't checking whether the passed in variable was null/undefined.

packages/opds-web-client/src/state.ts
```diff
@@ -18,8 +18,8 @@ export interface State {
     package, but it's available for other apps if they need to build the state
     for server-side rendering. */
 export default function buildInitialState(
-  collectionUrl: string,
-  bookUrl: string
+  collectionUrl?: string,
+  bookUrl?: string
```
kristojorg: For example, @adriana-alter , here you can imagine that the initial state could be built without a bookUrl - in fact unless the user is specifically requesting a /book/:bookUrl route, I believe there would be no bookUrl provided here, but it is marked as required. Which means that when you turn on strictNullChecks, I had to pass in a string at all times to this function.

kristojorg:
Here is some more background on that compiler flag: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html

For me, checking that variables are defined is possibly the single most useful thing that typescript can do

packages/opds-web-client/src/components/DownloadButton.tsx
```diff
@@ -75,7 +76,7 @@ export default class DownloadButton extends React.Component<
       return this.props.fulfill(this.props.url).then(blob => {
         download(
           blob,
-          this.generateFilename(this.props.title),
+          generateFilename(this.props.title, this.fileExtension()),
```
EdwinGuzman: This is fine but the reference to the original `generateFilename` function is still in this file and can be removed.

packages/opds-web-client/src/components/DownloadButton.tsx
```diff
@@ -124,9 +118,8 @@ export default class DownloadButton extends React.Component<
     ) {
       return "Read Online";
     }
-    let type = this.fileExtension()
-      .replace(".", "")
-      .toUpperCase();
+    let type = typeMap[this.mimeType()]?.name;
```
EdwinGuzman: Nice.

packages/opds-web-client/src/components/DownloadButton.tsx
```diff
@@ -107,14 +108,7 @@ export default class DownloadButton extends React.Component<
...
   fileExtension() {
-    return (
-      {
-        "application/epub+zip": ".epub",
-        "application/pdf": ".pdf",
-        "application/vnd.adobe.adept+xml": ".acsm",
-        "application/x-mobipocket-ebook": ".mobi"
-      }[this.mimeType()] || ""
-    );
+    return typeMap[this.mimeType()]?.extension ?? "";
```
EdwinGuzman: So, basically, if the left side is ever null or undefined, then and only then return `""`, right?

EdwinGuzman: For Typescript syntax that is new and/or is not already in the repo, do you mind writing a small comment about it? It'll make it easier to read.

packages/opds-web-client/src/components/__tests__/DownloadButton-test.tsx
```diff
@@ -43,15 +43,15 @@ describe("DownloadButton", () => {
   it("shows button", () => {
     let button = wrapper.find("button");
     expect(button.props().style).to.deep.equal(style);
-    expect(button.text()).to.equal("Download EPUB");
+    expect(button.text()).to.equal("Download ePub");
```
EdwinGuzman: I forgot if we talked about this before, but is it okay to switch to `"ePub"` now?

packages/opds-web-client/src/components/context/StoreContext.tsx
```diff
@@ -1,16 +1,16 @@
...
 import * as Redux from "redux";
 import { State } from "../../state";
 import AuthPlugin from "../../AuthPlugin";
 import buildStore from "../../store";
 import { PathForContext } from "./PathForContext";
 import BasicAuthPlugin from "../../BasicAuthPlugin";
 
-type OPDSStoreProps = {
-  children: React.ReactElement;
+export type OPDSStoreProps = {
```
EdwinGuzman: Was this exported for cpw?

packages/opds-web-client/src/utils/file.ts
```diff
@@ -0,0 +1,27 @@
...
+    extension: ".acsm",
+    name: "ACSM"
+  },
+  "application/x-mobipocket-ebook": {
+    extension: ".mobi",
+    name: "MOBI"
+  }
+};
+
+export const generateFilename = (str: string, extension: string) => {
```
EdwinGuzman: Can you write tests for this function now that it's a util function?
When you remove the `generateFilename` method from the DownloadButton component, a test should fail.

packages/opds-web-client/src/state.ts
adriana-alter: Makes sense; thank you!

packages/opds-web-client/src/components/__tests__/DownloadButton-test.tsx
kristojorg: Ah right, I meant to ask this question in my PR - I'm not sure if it okay or not. We can keep EPUB, I just see ePub around the internet most of the time

packages/opds-web-client/src/components/context/StoreContext.tsx
kristojorg: Yep

packages/opds-web-client/src/utils/file.ts
kristojorg: Yep on it

packages/opds-web-client/src/components/DownloadButton.tsx
kristojorg: Will do, and yes that's right. It's very similar to doing ` const blah = maybeDefined || fallback`, except that it only uses the fallback for `null` and `undefined`, while ` || ` will also fall back for falsy values (0, false, etc).

packages/opds-web-client/src/utils/file.ts
kristojorg: I didn't see any existing tests but I did write some of my own. Let me know if you're sufficient

kristojorg:
I also now have made a string literal type for the `OpenAccessLinksType` which is used to make sure that the `typeMap` for example is complete. It refines the type from string, to only a few strings that are possible. 

We should do the same for `fullfillmentLinks`, but I haven't gotten around to dealing with those. They should be the same I believe though, correct? If so, I'll use the same enum there and we can get that change into this PR.

packages/opds-web-client/src/components/__tests__/DownloadButton-test.tsx
EdwinGuzman: Yes, let's keep `EPUB` for now.

EdwinGuzman:
This looks good. I think the fulfillment links are slightly different but that doens't have to be in this PR.

packages/opds-web-client/src/components/__tests__/DownloadButton-test.tsx
kristojorg: Switched it back : )

kristojorg:
I don't have permission to do the actual merge, so I'll let y'all do that when you're ready.

kristojorg:
Ready to merge this?

EdwinGuzman:
Yes, merging now. Thanks.

PR #250, opened 2020-01-31
Converting DownloadButton to a functional component and using hooks for actions.

EdwinGuzman:
@kristojorg , can you review this and let me know what you think so far?

I set up some actions and context as you recommended. The idea was to try to remove functions props from the `Lane` component but then realized that it would work better if it was from the `DownloadButton` component. This was mostly to remove the`fulfillBook` and `indirectFulfillBook` action props from Lanes -> Lane -> Book.

It sort of seems like more boilerplate-ish code to just pull in the actions and get state from the store but maybe I'll like it more when I keep using it.

I get a lot of linter errors but not sure why or how to fix them. If you check https://travis-ci.com/NYPL-Simplified/opds-web-client/builds/146845160 you'll see what I mean.

### DIFF
--- a/packages/opds-web-client/src/components/Book.tsx
+++ b/packages/opds-web-client/src/components/Book.tsx
@@ -11,8 +11,6 @@ export interface BookProps {
   book: BookData;
   collectionUrl?: string;
   updateBook: (url: string) => Promise<BookData>;
-  fulfillBook: (url: string) => Promise<Blob>;
-  indirectFulfillBook: (url: string, type: string) => Promise<string>;
   isSignedIn?: boolean;
   epubReaderUrlTemplate?: (epubUrl: string) => string;
 }
@@ -202,8 +200,6 @@ export default class Book<P extends BookProps> extends React.Component<P, {}> {
           return (
             <DownloadButton
               key={`${link.url}-${index}`}
-              fulfill={this.props.fulfillBook}
-              indirectFulfill={this.props.indirectFulfillBook}
               url={link.url}
               mimeType={link.type}
               title={this.props.book.title}
@@ -228,8 +224,6 @@ export default class Book<P extends BookProps> extends React.Component<P, {}> {
           return (
             <DownloadButton
               key={`${link.url}-${index}`}
-              fulfill={this.props.fulfillBook}
-              indirectFulfill={this.props.indirectFulfillBook}
               url={link.url}
               mimeType={link.type}
               title={this.props.book.title}
--- a/packages/opds-web-client/src/components/Breadcrumbs.tsx
+++ b/packages/opds-web-client/src/components/Breadcrumbs.tsx
@@ -11,36 +11,37 @@ export interface ComputeBreadcrumbs {
   (collection: CollectionData, history: LinkData[]): LinkData[];
 }
 
-/** Shows a list of breadcrumbs links above a collection. */
-export default class Breadcrumbs extends React.Component<BreadcrumbsProps, {}> {
-  public static defaultProps: Partial<BreadcrumbsProps> = {
-    currentLink: false
-  };
-  render(): JSX.Element {
-    return (
-      <nav aria-label="breadcrumbs" role="navigation">
-        <ol className="breadcrumbs">
-          {this.props.links &&
-            this.props.links.map((link, i) => (
-              <li key={link.url} className="breadcrumb">
-                {i === this.props.links.length - 1 &&
-                !this.props.currentLink ? (
-                  <span>{link.text}</span>
-                ) : (
-                  <CatalogLink collectionUrl={link.url} bookUrl={null}>
-                    {link.text}
-                  </CatalogLink>
-                )}
-              </li>
-            ))}
-        </ol>
-      </nav>
-    );
-  }
-}
+/**
+ * Shows a list of breadcrumbs links above a collection.
+ **/
+const Breadcrumbs: React.FC<BreadcrumbsProps> = ({ links, currentLink }) => (
+  <nav aria-label="breadcrumbs" role="navigation">
+    <ol className="breadcrumbs">
+      {links &&
+        links.map((link, i) => (
+          <li key={link.url} className="breadcrumb">
+            {i === links.length - 1 && !currentLink ? (
+              <span>{link.text}</span>
+            ) : (
+              <CatalogLink collectionUrl={link.url} bookUrl={null}>
+                {link.text}
+              </CatalogLink>
+            )}
+          </li>
+        ))}
+    </ol>
+  </nav>
+);
+Breadcrumbs.defaultProps = {
+  currentLink: false
+};
 
-/** Computes breadcrumbs based on the browser history, with the current collection as the
-    final element. */
+export default Breadcrumbs;
+
+/**
+ * Computes breadcrumbs based on the browser history, with the current
+ * collection as the final element.
+ **/
 export function defaultComputeBreadcrumbs(
   collection: CollectionData,
   history: LinkData[]
@@ -57,9 +58,12 @@ export function defaultComputeBreadcrumbs(
   return links;
 }
 
-/** Computes breadcrumbs assuming that the OPDS feed is hierarchical - uses the catalog root
-    link, the parent of the current collection if it's not the root, and the current collection.
-    The OPDS spec doesn't require a hierarchy, so this may not make sense for some feeds. */
+/**
+ * Computes breadcrumbs assuming that the OPDS feed is hierarchical - uses
+ * the catalog root link, the parent of the current collection if it's not
+ * the root, and the current collection. The OPDS spec doesn't require a
+ * hierarchy, so this may not make sense for some feeds.
+ * */
 export function hierarchyComputeBreadcrumbs(
   collection: CollectionData,
   history: LinkData[],
--- a/packages/opds-web-client/src/components/Breadcrumbs.tsx
+++ b/packages/opds-web-client/src/components/Breadcrumbs.tsx
@@ -14,7 +14,10 @@ export interface ComputeBreadcrumbs {
 /**
  * Shows a list of breadcrumbs links above a collection.
  **/
-const Breadcrumbs: React.FC<BreadcrumbsProps> = ({ links, currentLink }) => (
+const Breadcrumbs: React.FC<BreadcrumbsProps> = ({
+  links,
+  currentLink = false
+}) => (
   <nav aria-label="breadcrumbs" role="navigation">
     <ol className="breadcrumbs">
       {links &&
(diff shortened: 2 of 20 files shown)


packages/opds-web-client/src/components/DownloadButton.tsx
```diff
@@ -1,132 +1,99 @@
...
+    downloadProps["onClick"] = fulfill;
+    downloadProps["className"] =
+      `${downloadProps["className"]} ` +
+      `download-${fileExtension().slice(1)}-button`;
   }
-}
+
+  return (
+    <span>
+      {React.createElement(
```
kristojorg: I would replace this with just a ternary render, and then you can also get rid of the if/else statement above. Something like this:
```js
<span>
    {isPlainLink ? 
        <a className="..." href={url} target="_blank" /** your other props */ />  : 
        <button /** props for the button version */ /> 
    }
</span>
```

kristojorg: That seems simpler to read and reason about what props are going where

packages/opds-web-client/src/components/DownloadButton.tsx
```diff
@@ -1,132 +1,99 @@
...
+    isPlainLink,
+    indirectType,
+    title,
+    type,
+    ...elementProps
+  } = props;
+  const { actions, dispatch } = useActions();
+  const fulfill = () => {
+    let dispatchFn;
+    let action;
```
kristojorg: I would personally just return these directly in the if statements:
```js
if (isIndirect()) {
    const action = ...
    return dispatch(action).then(...)
}
const action = ...
return dispatch(action).then(...)
```

I try to steer clear of having unbound variables that get dynamically bound to things on the whole in order to avoid unexpected undefined cases. Keep in mind that this package doesn't have `strictNullChecks` turned on in `tsconfig`, so typescript won't necessarily catch type errors where those variables don't get defined before they're used

packages/opds-web-client/src/components/__tests__/DownloadButton-test.tsx
```diff
@@ -61,71 +91,71 @@ describe("DownloadButton", () => {
...
-        done();
-      })
-      .catch(err => {
-        console.log(err);
-        throw err;
-      });
+    await button.props().onClick();
+    expect(downloadMock.getBlob()).to.equal("blob");
+    expect(downloadMock.getFilename()).to.equal("title.epub");
+    expect(downloadMock.getMimeType()).to.equal("application/epub+zip");
```
kristojorg: So much easier to read woah 😍

packages/opds-web-client/src/components/DownloadButton.tsx
EdwinGuzman: I initially tried this but the linter (maybe not Typescript) complained about it. This was the way to get around it. I'll double-check but did find this annoying to deal with.

packages/opds-web-client/src/components/Breadcrumbs.tsx
```diff
@@ -11,36 +11,37 @@ export interface ComputeBreadcrumbs {
...
+              <CatalogLink collectionUrl={link.url} bookUrl={null}>
+                {link.text}
+              </CatalogLink>
+            )}
+          </li>
+        ))}
+    </ol>
+  </nav>
+);
+Breadcrumbs.defaultProps = {
```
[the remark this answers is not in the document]
EdwinGuzman: Oh true. I usually do that but missed it this time.

packages/opds-web-client/src/components/__tests__/Book-test.tsx
```diff
@@ -319,18 +251,19 @@ describe("Book", () => {
...
+      let store = buildStore();
       wrapper = mount(
-        <Book
-          book={book}
-          updateBook={updateBook}
-          fulfillBook={stub()}
-          indirectFulfillBook={stub()}
-        />,
+        <Provider store={store}>
+          <ActionsProvider>
```
EdwinGuzman: @kristojorg  is there an alternative to using the providers here for mounting a component to test? I can see this expanding to many providers later on and having it cleaner to look at and easier to test would be great.

packages/opds-web-client/src/components/DownloadButton.tsx
kristojorg: Huh, yea let me know what lint setting or typescript setting complained. Should probably be removed as this would be better

packages/opds-web-client/src/components/__tests__/Book-test.tsx
kristojorg: I would suggest we rely on standard react composition or functions to remove that boilderplate. Would something like this help? 
```js
const AppContext = ({store, children, ...etc}) => (
<Provider store={store}>
    <ActionsProvider>
        {children}
    </ActionsProvider>
</Provider>
```

You then just wrap with that when testing. Alternatively you could do something like
```js
const render = (store, ...etc) => (
<Provider store={store}>
    <ActionsProvider>
        {children}
    </ActionsProvider>
</Provider>
```
as a standard function that you call. 

I'll do a little more research though to see if there are better ways to do this

EdwinGuzman: The first one looks good. For now, I did what's in this PR: https://github.com/NYPL-Simplified/opds-web-client/pull/250/files#diff-f67235cabd8fcea381a5d0791f5947d9R29-R35

packages/opds-web-client/src/components/DownloadButton.tsx
EdwinGuzman: Okay, so it's fine. What I couldn't do before was
```
return dispatch(actions.indirectFulfillBook(url, indirectType)).then(
```

packages/opds-web-client/src/components/Book.tsx
```diff
@@ -11,8 +11,6 @@ export interface BookProps {
   book: BookData;
   collectionUrl?: string;
   updateBook: (url: string) => Promise<BookData>;
-  fulfillBook: (url: string) => Promise<Blob>;
-  indirectFulfillBook: (url: string, type: string) => Promise<string>;
```
EdwinGuzman: @adriana-alter , as part of the redux and hook update, we don't need to pass down these actions as props down from the root to the leaf node. In this specific case, these two functions were being passed from 
```
Root.tsx -> BookDetails (which inherits from Book) -> DownloadButton.tsx
```

The two functions as props were not being as at all in BookDetails or Book so now that's why this is being removed here. Now, only DownloadButton knows about these actions and calls them. The code reduction is nice, but I do see how this makes DownloadButton less reusable _in other apps_, but works as intended in for this app.

packages/opds-web-client/src/components/Collection.tsx
```diff
@@ -93,8 +93,6 @@ export default class Collection extends React.Component<CollectionProps, {}> {
                 url={this.props.collection.url}
                 lanes={this.props.collection.lanes}
                 updateBook={this.props.updateBook}
-                fulfillBook={this.props.fulfillBook}
-                indirectFulfillBook={this.props.indirectFulfillBook}
```
EdwinGuzman: @adriana-alter same issue as before! This time it is:
```
Root.tsx -> Collection.tsx -> Lanes.tsx -> Lane.tsx -> Book.tsx -> DownloadButton.tsx
```

A long way to send these two actions for DownloadButton to use.

EdwinGuzman:
@aslagle I'm not sure if you want to be added a reviewer but I want you to be aware of these changes that we are doing for all the apps. I think this is resolving an issue of an older ticket that involved updating Redux so each child connects to the store rather than the app connecting to it at the root.

packages/opds-web-client/src/components/Book.tsx
kristojorg: I've been thinking about this too, and it's true that `useSelector` couples a component to the redux implementation within the app, while `connect` is creating a wrapper component that deals with the logic outside of the presentation component, which in this case would be the button. 

I think there are valid use cases for both `useSelector` and `connect`. In this case, if OPDS was meant to be a UI component library, it would make sense to have a series of UI-only components like `DownloadButton`. But in my experience, it's the logic in OPDS that is useful, not usually the UI components. I think that's the same for simply-e and circulation-web. So the goal shouldn't be to make all the UI components reusable, but the logic reusable:
```js
const MyDownloadButton = ({book}) => {
    // gets fulfill, dispatch, etc internally
    const {label, onClick, ...etc} = useDownloadButton(book);
    const myOnClick = () => {
        // something custom
        onClick();
    }
    return (
        <CustomButtonUI onClick={myOnClick} {...etc}>{label}</CustomButtonUI>
    );
}
```
So yes, `connect` is better at decoupling a component from redux, but at the cost of some boilerplate and indirection. We should consider what decoupling is actually useful for consumers of OPDS, and in my experience it isn't making it so the UI components can be used without a redux. The first step in making any of the logic reusable is getting rid of that `mergeRootProps` file, so I think this is definitely a step in the right direction.

Just my 2 cents after working in CPW 🙂

packages/opds-web-client/src/components/DownloadButton.tsx
kristojorg: I would personally still get rid of the `else` block and the `let action`, instead binding action directly. But it's not a huge deal either way

packages/opds-web-client/src/components/DownloadButton.tsx
```diff
@@ -1,119 +1,93 @@
...
+  const fileExtension = (mimeTypeValue: string) =>
     // this ?? syntax is similar to x || y, except that it will only
     // fall back if the predicate is undefined or null, not if it
     // is falsy (false, 0, etc).
-    return typeMap[this.mimeType()]?.extension ?? "";
-  }
-
-  downloadLabel() {
+    typeMap[mimeTypeValue]?.extension ?? "";
+  const downloadLabel = () => {
```
kristojorg: Can also be a variable I think

packages/opds-web-client/src/components/__tests__/Book-test.tsx
kristojorg: Nice yea, I think that's basically the second one. Looks good to me

aslagle:
I haven't kept up with react/redux at all recently so I wouldn't be a good reviewer, but it's great that you're finally working on this 🙌

packages/opds-web-client/src/components/Book.tsx
EdwinGuzman: Yes, that's nicely worded. Those buttons aren't necessarily used in the admin as much as the logic and type interfaces, so it shouldn't be a big deal regarding coupling and reusability. I also wanted to make sure I was able to describe the trade-off that is happening due to this refactoring, so thanks for the verification and longer explanation.

EdwinGuzman:
Reviewed this with @adriana-alter in person 🙌

PR #251, opened 2020-02-06
Strict null checks

kristojorg:
I am requesting that this be merged into my own beta branch, but I want to get everyone's eyes on this first.

This enables typescripts [`strictNullChecks`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#--strictnullchecks) flag. The change is illustrated below:
```js
// this used to pass
const something: string = null;
// so typescript wouldn't catch this TypeError. Which is dangerous!
something.toUppercase();

function blah(var: string){...}
// so did this
blah();

// now we have to type things explicitly with null/undefined if we want to allow those types
const something: number = null;
//     ^^^ Error: can't assign null to number

// this is how we would do that instead:
const something: number | null = null;
```
In the end, this means that I had to pop around the repo and explicitly type a bunch of things as unions with `null` and `undefined`. In some cases it was a bit more complex than that, and I had to add new checks to make sure something wasn't undefined or null. 

This should make the code base far more stable. The most common production error of any codebase I've ever worked on is accessing a property on `undefined` or `null`, and this should virtually eliminate that. Plus it will make development easier as it will warn you as you're typing if you're doing something wrong. 

Let me know what you guys think

### DIFF
--- a/packages/opds-web-client/src/OPDSDataAdapter.ts
+++ b/packages/opds-web-client/src/OPDSDataAdapter.ts
@@ -173,17 +173,21 @@ export function entryToBook(entry: OPDSEntry, feedUrl: string): BookData {
   };
 }
 
-function entryToLink(entry: OPDSEntry, feedUrl: string): LinkData {
-  let href: string;
+function entryToLink(entry: OPDSEntry, feedUrl: string): LinkData | null {
   let links = entry.links;
   if (links.length > 0) {
-    href = resolve(feedUrl, links[0].href);
+    const href = resolve(feedUrl, links[0].href);
+    return {
+      id: entry.id,
+      text: entry.title,
+      url: href
+    };
   }
-  return <LinkData>{
-    id: entry.id,
-    text: entry.title,
-    url: href
-  };
+  console.error(
+    "Attempting to create Link with undefined url. entry is: ",
+    entry
+  );
+  return null;
 }
 
 function dedupeBooks(books: BookData[]): BookData[] {
@@ -221,7 +225,7 @@ function formatDate(inputDate: string): string {
   return `${month} ${day}, ${year}`;
 }
 
-function OPDSLinkToLinkData(feedUrl, link: OPDSLink = null) {
+function OPDSLinkToLinkData(feedUrl, link: OPDSLink = null): LinkData | null {
   if (!link || !link.href) {
     return null;
   }
@@ -245,20 +249,24 @@ export function feedToCollection(
   let books: BookData[] = [];
   let navigationLinks: LinkData[] = [];
   let lanes: LaneData[] = [];
-  let laneTitles = [];
-  let laneIndex = [];
+  let laneTitles: any[] = [];
+  let laneIndex: {
+    title: any;
+    url: string;
+    books: BookData[];
+  }[] = [];
   let facetGroups: FacetGroupData[] = [];
-  let search: SearchData;
-  let nextPageUrl: string;
+  let search: SearchData | undefined = undefined;
+  let nextPageUrl: string | undefined = undefined;
   let catalogRootLink: OPDSLink;
   let parentLink: OPDSLink;
-  let shelfUrl: string;
+  let shelfUrl: string | undefined = undefined;
   let links: OPDSLink[] = [];
 
   feed.entries.forEach(entry => {
     if (feed instanceof AcquisitionFeed) {
       let book = entryToBook(entry, feedUrl);
-      let collectionLink: OPDSCollectionLink = entry.links.find(
+      const collectionLink: OPDSCollectionLink = entry.links.find(
         link => link instanceof OPDSCollectionLink
       );
       if (collectionLink) {
@@ -280,7 +288,7 @@ export function feedToCollection(
       }
     } else {
       let link = entryToLink(entry, feedUrl);
-      navigationLinks.push(link);
+      if (link) navigationLinks.push(link);
     }
   });
 
@@ -291,7 +299,7 @@ export function feedToCollection(
     return result;
   }, lanes);
 
-  let facetLinks = [];
+  let facetLinks: OPDSFacetLink[] = [];
   if (feed.links) {
     facetLinks = feed.links.filter(link => {
       return link instanceof OPDSFacetLink;
@@ -331,7 +339,7 @@ export function feedToCollection(
     let href = resolve(feedUrl, link.href);
     let active = link.activeFacet;
     let facet = { label, href, active };
-    let newResult = [];
+    let newResult: any[] = [];
     let foundGroup = false;
     result.forEach(group => {
       if (group.label === groupLabel) {
@@ -358,7 +366,13 @@ export function feedToCollection(
   collection.catalogRootLink = OPDSLinkToLinkData(feedUrl, catalogRootLink);
   collection.parentLink = OPDSLinkToLinkData(feedUrl, parentLink);
   collection.shelfUrl = shelfUrl;
-  collection.links = links.map(link => OPDSLinkToLinkData(feedUrl, link));
+  function notNull<T>(value: T | null | undefined): value is T {
+    return value !== null && value !== undefined;
+  }
+  collection.links = links
+    .map(link => OPDSLinkToLinkData(feedUrl, link))
+    // we have to filter out the null values in order for typescript to accept this
+    .filter(notNull);
   collection.raw = feed.unparsed;
   Object.freeze(collection);
   return collection;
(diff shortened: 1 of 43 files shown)


kristojorg:
Note: I have a few tests to fix... will get on that ASAP

packages/opds-web-client/src/OPDSDataAdapter.ts
```diff
@@ -173,17 +173,21 @@ export function entryToBook(entry: OPDSEntry, feedUrl: string): BookData {
...
+      text: entry.title,
+      url: href
+    };
   }
-  return <LinkData>{
-    id: entry.id,
-    text: entry.title,
-    url: href
-  };
+  console.error(
```
kristojorg: I added this. It used to return an object an undefined `href`, but that didn't seem correct. Let me know what you think here @EdwinGuzman

packages/opds-web-client/src/OPDSDataAdapter.ts
```diff
@@ -245,20 +249,24 @@ export function feedToCollection(
...
+  let laneTitles: any[] = [];
+  let laneIndex: {
+    title: any;
+    url: string;
+    books: BookData[];
+  }[] = [];
   let facetGroups: FacetGroupData[] = [];
-  let search: SearchData;
-  let nextPageUrl: string;
+  let search: SearchData | undefined = undefined;
```
kristojorg: We now have to initialize variables even if they're initialized as `undefined`

packages/opds-web-client/src/__tests__/OPDSDataAdaptor-test.ts
```diff
@@ -102,32 +102,32 @@ describe("OPDSDataAdapter", () => {
...
     expect(book.title).to.equal(entry.title);
-    expect(book.authors.length).to.equal(2);
-    expect(book.authors[0]).to.equal(entry.authors[0].name);
-    expect(book.authors[1]).to.equal(entry.authors[1].name);
-    expect(book.contributors.length).to.equal(1);
-    expect(book.contributors[0]).to.equal(entry.contributors[0].name);
-    expect(book.series.name).to.equal(entry.series.name);
-    expect(book.series.position).to.equal(entry.series.position);
+    expect(book.authors?.length).to.equal(2);
+    expect(book.authors?.[0]).to.equal(entry.authors[0].name);
```
kristojorg: You can see that with `strictNullChecks` enabled, typescript is now catching all of these previously possibly undefined property access errors, which is why I have to use optional chaining to make sure I'm not going to get an error at runtime. 

Not super helpful in these tests, but those kind of errors would crash the app in production

packages/opds-web-client/src/components/Collection.tsx
```diff
@@ -210,13 +210,15 @@ export default class Collection extends React.Component<CollectionProps, {}> {
...
+      !!this.props.collection.nextPageUrl
     );
   }
 
   fetch() {
-    if (this.canFetch()) {
-      this.props.fetchPage(this.props.collection.nextPageUrl);
+    // had to move this typeguard into here for typescript to
+    // believe the nextPageUrl was defined.
+    if (this.canFetch() && this.props.collection.nextPageUrl) {
```
kristojorg: One annoying thing about typescript is that you can't save a typeguard into a variable:
```js
function(a: string | null) {
  const isString = !!string;
  if (isString){
    return a.toUppercase();
}
// you'll get an error at toUppercase above, because typescript doesn't realize the type has been narrowed by isString
if(a) {
  return a.toUppercase();
} 
// the typeguard has to be inline for it not to error
```

packages/opds-web-client/src/components/DownloadButton.tsx
```diff
@@ -68,15 +68,20 @@ export default class DownloadButton extends React.Component<
...
     } else {
-      return this.props.fulfill(this.props.url).then(blob => {
+      return this.props.fulfill?.(this.props.url).then(blob => {
         download(
           blob,
-          generateFilename(this.props.title, this.fileExtension()),
+          generateFilename(
+            // the following is meant to smoothly handle the case where
+            // this.props.title is undefined or null
+            this.props.title ?? "untitled",
```
kristojorg: I wasn't totally sure what to do in this situation. It was a case that previously could have happened but wasn't really being handled.

packages/opds-web-client/CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 ## Changelog
 
+### v0.3.4
```
EdwinGuzman: This is fine but it may not match one-to-one with the changes that can potentially be added in `master`. Not a big deal but just something to keep in mind.

packages/opds-web-client/src/components/DownloadButton.tsx
EdwinGuzman: I'm not sure if this is the safest option but seems okay for now.

packages/opds-web-client/src/components/Root.tsx
```diff
@@ -323,28 +340,11 @@ export class Root extends React.Component<RootProps, RootState> {
...
-                  setPreference={this.props.setPreference}
-                />
-              )
-            ) : null}
+                )
+              ) : null)}
           </div>
         </main>
-        {showFooter && (
+        {Footer && this.props.collectionData && (
```
EdwinGuzman: Any reason why the variable was removed in favor for this? It seems `showFooter` was the same thing.

packages/opds-web-client/src/OPDSDataAdapter.ts
EdwinGuzman: I can't say I know for sure if this is right (as in, an acceptable fallback for the data structure).

EdwinGuzman:
I'm going to look more into this to better learn it. For initially going through the PR, it seems like a lot of minor things were updated to pass this null check. I can see why it's worth it but it makes me think that perhaps it would be better to get the low-level details about the data correct, so to not have `| null` (or similar) in many places.

packages/opds-web-client/src/components/Root.tsx
kristojorg: Yea, my comment somewhere explains it in detail. Typescript can't store a "type guard" in a variable, so `showFooter` worked, but typescript didn't understand that that variable meant both `Footer` and `this.props.collectionData` were defined. This was always the case, it's just that before `strictNullChecks`, typescript was fine with accessing properties on possibly undefined vars. 

So long story short, in order to allow typescript to refine the type so that it knows those vars are defined, we have to perform the check inline, or write a type guard function that does it. I could have turned `showFooter` into one of those functions, but decided this was simpler. 

Hopefully one day we will be able to type a boolean like this:
```js
const showFooter: Footer is JSX.Element = !!Footer;
```

kristojorg:
Yea I was thinking about that... it would make sense for sure to get the types right, but this isn't adding flexibility to the types... it's just being explicit about when a type is nullable. In the past that was implicit in every type definition. 

I see how it's a lot of changes, but it makes the code much more explicit in my mind, and you can then go through and start removing `| null` unions as you tighten up the code and handle `null` cases higher up the call stack. 

It looks like it's making the code less safe because I'm adding `null` all over, but in reality it's still getting much safer because those vars were already nullable, and there are lots of vars I didn't add `null` to that used to be implicitly nullable which now are not

packages/opds-web-client/CHANGELOG.md
kristojorg: Okay yep I can change to whatever is needed

packages/opds-web-client/src/components/DownloadButton.tsx
kristojorg: Yea I wasn't sure what to do here. Seemed like we should do something... in the past this would have caused a type error and the app would have crashed, so this is better than that at least.

packages/opds-web-client/src/components/Collection.tsx
kristojorg: @EdwinGuzman re your commend on the type guard I inlined that used to be in the `showFooter` var.

packages/opds-web-client/src/OPDSDataAdapter.ts
kristojorg: Yea in the past, it would have returned an object with an undefined `href`, so maybe I should just preserve that functionality I suppose for now?

EdwinGuzman: Yeah, I think for now just to be safe until we know for sure later.

kristojorg:
Okay I'll go ahead and fix the tests then. I agree it's uglier, but at least now you know when things might be `null` / `undefined`.

kristojorg:
Okay fixed the tests, merging this into my beta branch.

PR #255, opened 2020-02-13
beta

kristojorg:
Okay I realize now I should have branched my `strict-null-checks` branch off of master and made that a separate PR so that this one was more clear. In any case, this is what it accomplishes:
1. strict null checks is turned on 
2. adds kepub support (I found these files in the catalog I was working with in circulation-patron-web).
3. extracts some logic from classes into functions so that it can be imported into circulation-patron-web.


kristojorg:
For simplicity of reviewing (since you've already reviewed the strict null checks stuff), maybe try just clicking on individual commits to see them in isolation.

EdwinGuzman:
The new version should be 0.3.5 and some files are out of date.

packages/opds-web-client/src/components/Root.tsx
```diff
@@ -138,14 +138,13 @@ export class Root extends React.Component<RootProps, RootState> {
     );
 
     let showCollection = this.props.collectionData && !this.props.bookData;
-    let showBook = this.props.bookData;
+    const showBook = !!this.props.bookData;
```
EdwinGuzman: This was updated but on line 265 it was replaced by `this.props.bookData`. This is still being used for `showCollectionContainer` on line 150. Shouldn't that be updated as well?

packages/opds-web-client/src/utils/book.ts
```diff
@@ -0,0 +1,21 @@
...
+export function bookIsReserved(book: BookData) {
+  return book.availability?.status === "reserved";
+}
+
+export function bookIsReady(book: BookData) {
+  return book.availability?.status === "ready";
+}
+
+export function bookIsBorrowed(book: BookData) {
+  return book.fulfillmentLinks?.length ?? 0 > 0;
```
EdwinGuzman: Why is the second clause `0 > 0`? This means that if and only if `length` is null or undefined, then check that 0 is greater than 0?

packages/opds-web-client/src/reducers/history.ts
```diff
@@ -28,8 +29,11 @@ function newRootIsNotOldRoot(
   newCollection: CollectionData,
   oldCollection: CollectionData | undefined | null
 ): boolean {
-  return (
-    newCollection?.catalogRootLink?.url !== oldCollection?.catalogRootLink?.url
+  return !!(
+    oldCollection &&
+    newCollection.catalogRootLink &&
+    oldCollection.catalogRootLink &&
+    newCollection.catalogRootLink.url !== oldCollection.catalogRootLink.url
```
EdwinGuzman: Why didn't the previous implementation work?

packages/opds-web-client/src/interfaces.ts
```diff
@@ -192,3 +192,9 @@ export interface BasicAuthMethod extends AuthMethod {
     password: string;
   };
 }
+
+type PickAndRequire<T, K extends keyof T> = { [P in K]-?: NonNullable<T[P]> };
```
EdwinGuzman: Can you add a comment about what this is doing? Looks like just making sure all the properties from K are present, and also not null?

packages/opds-web-client/src/utils/book.ts
kristojorg: Oh shoot you're right I missed a parenthesis in there

packages/opds-web-client/src/reducers/history.ts
kristojorg: It's a slight thing but I found that it actually changed the semantics in this case. It's because the version with `&&` actually checks both that they are defined, and that they are not equal, whereas mine only checks that they are not equal... so if they are both undefined then they'll be equal, which isn't what we want in this case (It caused a test failure)

packages/opds-web-client/src/interfaces.ts
kristojorg: Yep will do. It allows you to pick which properties you want to force to be required: 
```js
type BookDataWithCopies = RequireKeys<BookData, "copies">;
```

packages/opds-web-client/src/components/Root.tsx
kristojorg: We can replace it and get rid of `showBook`, but it's not necessary. The reason it was necessary to replace `showBook` in that instance was because the following code then depended on `this.props.bookData` being of type `BookData`, and typescript would error unless it could narrow the type to that... whereas in this case the `showCollectionContainer` is used to control the display of some UI which doesn't depend on `this.props.bookData`, so there is no typescript error. In the end, the functionality of the code is the same whether you use `showBook` or `this.props.bookData`, it's just that typescript can't be sure that `this.props.bookData` is actually defined if you use `showBook`... does that make any sense?

packages/opds-web-client/src/interfaces.ts
kristojorg: Actually, `PickAndRequire` was only really usable in the context of `RequireKeys`, so I just inlined that type into `RequireKeys`, which does what I explained above

kristojorg:
Let me know if you're ready to approve this or if you spot anything else @EdwinGuzman

packages/opds-web-client/src/components/Root.tsx
```diff
@@ -372,15 +381,16 @@ export class Root extends React.Component<RootProps, RootState> {
     if (authError) {
       this.setState({ authError });
     } else if (this.props.collectionUrl || this.props.bookUrl) {
+      console.log("FETCH", this.props.collectionUrl, this.props.bookUrl);
```
EdwinGuzman: This should be removed.

packages/opds-web-client/src/components/__tests__/Root-test.tsx
```diff
@@ -38,17 +63,38 @@ const setCollectionAndBookPromise = new Promise((resolve, reject) => {
...
   it("shows skip navigation link", () => {
-    let wrapper = shallow(<Root />);
+    let wrapper = shallow(
+      <Root
+        fulfillBook={fulfillBook}
+        indirectFulfillBook={indirectFulfillBook}
+        updateBook={updateBook}
+        setPreference={setPreference}
+      />
+    );
```
EdwinGuzman: I think I missed this but why are these props not all mandatory?

EdwinGuzman:
It looks good but I just wanted to review it all once again and also test it. As far as I can tell, all the existing functionality works the same as before. Just a minor thing and one question.

packages/opds-web-client/src/components/CatalogLink.tsx
```diff
@@ -38,9 +38,9 @@ export default class CatalogLink extends React.Component<CatalogLinkProps, {}> {
   }
 
   render(): JSX.Element {
-    let { collectionUrl, bookUrl, ...props } = this.props;
-    collectionUrl = collectionUrl || null;
-    bookUrl = bookUrl || null;
+    let { collectionUrl = null, bookUrl = null, ...props } = this.props;
+    // collectionUrl = collectionUrl || null;
+    // bookUrl = bookUrl || null;
```
EdwinGuzman: This can also be removed.

packages/opds-web-client/src/components/__tests__/Root-test.tsx
kristojorg: I assume you mean "now all mandatory"? I made them required because `Root` was implicitly expecting them to be there and this seemed more logical than handling cases where they weren't defined. If you think they should be optional, I can make them that way but then we'll have to figure out how `Root` should work if they aren't defined, which I'm not really sure what the expected behavior would be.

packages/opds-web-client/src/components/Root.tsx
kristojorg: whoops !

kristojorg:
Ok made those changes! 👍

EdwinGuzman:
Also tested this branch on `circulation-web` and it's working well.

PR #256, opened 2020-02-19
Download button

kristojorg:
This PR extracts the logic that was previously in `DownloadButton` into a `useDownloadButton` hook that can be imported by CPW. This patterns allows CPW to use the same download logic as this repo while rendering a different visual element. 

I also added the `fetcher` to the context passed down by `ActionsContext` because it is necessary sometimes to access and use it directly. I am now importing the same `ActionsContext` and using it in CPW. Previously I had created my own (before OPDS had this).


packages/opds-web-client/src/hooks/useDownloadButton.ts
```diff
@@ -0,0 +1,95 @@
+import * as React from "react";
```
EdwinGuzman: There should be tests for this hook.

packages/opds-web-client/src/components/context/ActionsContext.tsx
```diff
@@ -5,16 +5,25 @@ import DataFetcher from "../../DataFetcher";
 import { adapter } from "../../OPDSDataAdapter";
 
 // The main context for this app's Actions.
-export const ActionsContext = React.createContext<ActionsCreator | undefined>(
+type ActionsContextType =
+  | { fetcher: DataFetcher; actions: ActionsCreator }
```
EdwinGuzman: Do you need this first pipe?

packages/opds-web-client/src/components/__tests__/Book-test.tsx
```diff
@@ -277,12 +277,13 @@ describe("Book", () => {
       let epubButton = buttons.at(0);
       let mobiButton = buttons.at(1);
 
-      expect(epubButton.props().url).to.equal("secrets.epub");
-      expect(epubButton.props().mimeType).to.equal("application/epub+zip");
+      console.log(epubButton.props().link.type);
```
EdwinGuzman: extra console

packages/opds-web-client/src/components/DownloadButton.tsx
```diff
@@ -1,89 +1,41 @@
...
+import { MediaLink, FulfillmentLink } from "./../interfaces";
 
-export interface DownloadButtonProps extends React.HTMLProps<{}> {
-  url: string;
-  mimeType: string;
-  isPlainLink?: boolean;
-  title?: string;
-  indirectType?: string;
-}
+export type DownloadButtonProps = React.ComponentProps<"button"> &
```
EdwinGuzman: Why the switch from `interface` to `type`?

packages/opds-web-client/src/hooks/useDownloadButton.ts
EdwinGuzman: I know these are indirectly tested in components, but explicit tests would be nice.

packages/opds-web-client/src/hooks/useDownloadButton.ts
```diff
@@ -0,0 +1,95 @@
...
+ * We use typescript function overloads to show that if you pass in a link
+ * you are guaranteed to get the download button details in an object. If
+ * you might pass in undefined, you might get null
+ */
+export default function useDownloadButton(
+  link: MediaLink | FulfillmentLink,
+  title: string
+): DownloadDetails;
+export default function useDownloadButton(
+  link: MediaLink | FulfillmentLink | undefined,
```
EdwinGuzman: Would this work if you did `link: undefined,` for this function declaration?

packages/opds-web-client/src/components/__tests__/DownloadButton-test.tsx
```diff
@@ -111,34 +116,31 @@ describe("DownloadButton", () => {
...
     expect(indirectFulfill.callCount).to.equal(1);
     expect(indirectFulfill.args[0][0]).to.equal("download url");
-    expect(indirectFulfill.args[0][1]).to.equal(streamingType);
+    expect(indirectFulfill.args[0][1]).to.equal(streamingLink.indirectType);
   });
 
   it("fulfills ACSM-based indirect links", async () => {
+    const ascmLink: FulfillmentLink = {
+      url: "download url",
+      type: "application/vnd.adobe.adept+xml",
```
EdwinGuzman: The original value for this was `"vnd.adobe/adept+xml"` and it should stay the same. The mimeType function can also be tested separately for this.

EdwinGuzman:
Looks good but I think the new hook may be missing tests and just general questions.

EdwinGuzman:
@adriana-alter take a look please!

packages/opds-web-client/src/app.tsx
```diff
@@ -32,7 +32,7 @@ const OPDSCatalogRouterHandler = config => {
       };
       return (
         <PathForProvider pathFor={config.pathFor}>
-          <ActionsProvider>
+          <ActionsProvider proxyUrl={config.proxyUrl}>
```
adriana-alter: What's the proxyUrl?

packages/opds-web-client/src/components/Book.tsx
```diff
@@ -203,11 +202,9 @@ export default class Book<P extends BookProps> extends React.Component<P, {}> {
         links.push(
           <DownloadButton
             key={`${link.url}-${index}`}
-            url={link.url}
-            mimeType={link.type}
+            link={link}
```
adriana-alter: Why don't we need the `mimeType` anymore?

adriana-alter:
This looks pretty good to me, but I'd also like to know what's up with the switch from `interface` to `type`!

packages/opds-web-client/src/components/context/ActionsContext.tsx
kristojorg: Yea that's how prettier formats it when you have a multi-line union. Makes it easier to read especially when you get over two unions.

packages/opds-web-client/src/components/DownloadButton.tsx
kristojorg: Was easier for me to union with multiple types. I don't think inheriting multiple interfaces works the same as this. I find types map better to modern JS than interfaces, since they're not using inheritance when it isn't necessary.

packages/opds-web-client/src/hooks/useDownloadButton.ts
kristojorg: For the declaration or for when you call the function? It allows you to call the function with a value that has type `MediaLink | FulfillmentLink | undefined`, but then it may return `null` (if in fact the value turns out to be `undefined`. However, if you pass it a value that is guaranteed to be `MediaLink | FulfillmentLink`, you are guaranteed to get back `DownloadDetails`. Does that make sense?

packages/opds-web-client/src/components/__tests__/DownloadButton-test.tsx
kristojorg: Okay, makes sense. The `MediaLink` type doesn't have that type in the enum: https://github.com/NYPL-Simplified/opds-web-client/blob/b66c88f2df2dc2e564b813137b2526da7ad53414/packages/opds-web-client/src/interfaces.ts#L3

I will add it, but that then requires that I add an entry for it in the `typeMap` found here: 
https://github.com/NYPL-Simplified/opds-web-client/blob/b66c88f2df2dc2e564b813137b2526da7ad53414/packages/opds-web-client/src/utils/file.ts#L10

Do you know what the extension and format name should be? For now I will use `.acsm` and `ACSM` for now.

packages/opds-web-client/src/app.tsx
kristojorg: The proxyUrl is something that I didn't build but is an option in the `DataFetcher` which is instantiated in `ActionsContextProvider`, so I passed it through to there

packages/opds-web-client/src/components/__tests__/DownloadButton-test.tsx
EdwinGuzman: Not sure but `.acsm` and `ACSM` is fine for now.

kristojorg:
Okay I've added tests for `useDownloadButton`. In order to do that I needed to update some of the testing packages like `sinon`, which subsequently required updating some of the tests to match the changed APIs of those packages. Should be all working now. Let me know what you think and if this is ready to be merged

kristojorg:
Looks like we need to update the node version. Do you know how to do that @EdwinGuzman ?

EdwinGuzman:
I see. In the Travis yml file, you can update the node version from 8 to 10. That should fix it.

EdwinGuzman:
Everything else looks good, thanks for updating the tests!

kristojorg:
@EdwinGuzman do you see any reason not to update to the latest node version (13.11.0)?

kristojorg:
Ok I've updated the node version to 13 (latest stable) and everything seems to be passing. Let me know if I should do anything else before merging this.

packages/opds-web-client/src/hooks/__tests__/useDownloadButton-test.ts
```diff
@@ -105,19 +106,76 @@ describe("useDownloadButton", () => {
...
+
+    const { result } = renderHook(
+      () => useDownloadButton(pdfMediaLink, "book-title"),
+      {
+        wrapper
+      }
+    );
+
+    // call fulfill
+    await result.current.fulfill();
```
EdwinGuzman: Is `current` part of the testing library API?

kristojorg: yea, it stores the result of the hook in a ref, so the way to access it is `result.current`, and you can't destructure the result, since that creates a local copy instead of maintaining reference to the ref variable.

kristojorg:
Okay I put v13 in the Travis config, should I make a new PR to bump it back down to 10?

PR #257, opened 2020-03-31
fix server-rendering jsdom code

kristojorg:
I upgraded the `jsdom` version in my `useDownloadButton` PR, and updated the code appropriately for testing, but I didn't see this use of `jsdom` for server rendering previously. I've updated it to the new `jsdom` api. Tests didn't catch this because that code block is uncovered (window is always defined in tests via the test `jsdom` env)

### DIFF
--- a/packages/opds-web-client/CHANGELOG.md
+++ b/packages/opds-web-client/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## Changelog
 
+### v0.3.7
+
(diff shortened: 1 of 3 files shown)


kristojorg:
I didn't know it could be used that way either, but have updated the changelog and version for ya

EdwinGuzman:
Just published a new version!

PR #258, opened 2020-04-07
Book medium

kristojorg:
This extracts the logic to `getMedium` and `getMediumSVG` from the `<Book />` component so it can be imported into CPW and used there. 

This does technically introduce a breaking change in that any component inheriting from `Book` will no longer have access to the `getMedium` or `getMediumSVG` methods. It appears `circulation-web` [is depending on these methods](https://github.com/NYPL-Simplified/circulation-web/blob/9c0a7aa5e40c6a688c71f758bbad7f66e3e3f49e/src/components/BookEditForm.tsx#L66). I can make a corresponding PR there to update to the new API. I marked this as a patch change but technically is should be a major change, would you like me to move us to 1.0.0?

Let me know what you think.

I also added the default prettier config to `package.json` to remove inconsistencies caused by editor plugins like the one I use in vscode changing their defaults. Just makes our config explicit so it won't change without our approval.


EdwinGuzman:
How about `0.4.0` for the version bump. The PR looks good, but yes, please make a respective PR to `circulation` with this update. I'm assuming you'll pull this into the current `circulation-patron-web` PR?

kristojorg:
Sounds good and yes this is so I can bring this in to CPW. I've updated the version, so if we get this merged and published I'll send a PR to `circulation-web` with the updated `opds-web-client` dep

EdwinGuzman:
For something like `opds-web-client` to `circulation-web`, I find that creating both PRs at the same time useful in case I need to update something in OWC. That way, a new and small version isn't published for something minor that was overlooked. But, is the issue that npm link won't work well for you so you want to wait until it is published?

kristojorg:
I can make a PR to `circulation-web` but I will be making changes that don't work yet since changing `this.getMedium` to the imported `getMedium` will break until the new `opds-web-client` version is published and installed. I just didn't want to push a PR that would cause something to break, but I suppose I could make a draft PR. In the end, `opds-web-client` still needs to be updated first though.

EdwinGuzman:
That's fine. I can locally test it with npm link and update the circulation-web PR if necessary.

packages/opds-web-client/src/components/Book.tsx
```diff
@@ -3,13 +3,15 @@ import CatalogLink from "./CatalogLink";
 import BookCover from "./BookCover";
 import BorrowButton from "./BorrowButton";
 import DownloadButton from "./DownloadButton";
-import { BookData, FulfillmentLink } from "../interfaces";
+import { BookData, FulfillmentLink, BookMedium } from "../interfaces";
 import { AudioHeadphoneIcon, BookIcon } from "@nypl/dgx-svg-icons";
```
EdwinGuzman, line 7: You don't need these in this file anymore.

EdwinGuzman:
Just one small change but everything looks good.

packages/opds-web-client/src/components/Book.tsx
kristojorg: Done

kristojorg:
Alright I removed the unused imports : )

PR #260, opened 2020-04-17
Audiobook support

kristojorg:
Adds audiobook support to the `typeMap` and `MediaType`.

### DIFF
--- a/packages/opds-web-client/src/utils/file.ts
+++ b/packages/opds-web-client/src/utils/file.ts
@@ -44,6 +44,10 @@ export const typeMap: TypeMap = {
   "text/html;profile=http://librarysimplified.org/terms/profiles/streaming-media": {
     extension: "",
     name: "streaming-media"
+  },
+  "application/audiobook+json": {
+    extension: ".audiobook",
+    name: "Audiobook"
   }
 };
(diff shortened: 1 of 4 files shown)


packages/opds-web-client/src/utils/file.ts
```diff
@@ -44,6 +44,10 @@ export const typeMap: TypeMap = {
   "text/html;profile=http://librarysimplified.org/terms/profiles/streaming-media": {
     extension: "",
     name: "streaming-media"
+  },
+  "application/audiobook+json": {
+    extension: ".audiobook",
```
EdwinGuzman, line 49: @leonardr is this right?

leonardr: I think `.audiobook-manifest` is more accurate. That's what we use in core/model/constants.py and that's what the example has in https://github.com/readium/webpub-manifest/blob/master/extensions/audiobook.md.

kristojorg:
I believe we've decided to leave this as `.audiobook`. In which case, is there anything blocking this merge?

kristojorg:
Do you mind merging? I'm not authorized here

PR #265, opened 2020-06-18
add streaming media flag

kristojorg:
Simply passes a flag from the `useDownloadButton` hook on whether the media is streaming or not.


EdwinGuzman:
Just needs a version bump.

kristojorg:
Sweet, did that! If you don't mind merging/publishing when you have a chance that would be appreciated : )

PR #266, opened 2020-07-10
Fix auth 

kristojorg:
Previously, when someone would log out, we would clear the cookie and then _not_ set an auth header on the subsequent request. This would result in the browser adding out Basic Auth header from its cache, which would result in continued logged-in requests until the user closed the tab/window or otherwise cleared their browser cache. 

To fix this, I made it so we always set an `Authorization` header, but if there are no credentials present we set it to an empty string, which results in a proper `401` response from authenticated endpoints.

### DIFF
--- a/packages/opds-web-client/CHANGELOG.md
+++ b/packages/opds-web-client/CHANGELOG.md
@@ -2,7 +2,8 @@
 
 ### v0.4.5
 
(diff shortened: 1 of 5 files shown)


kristojorg:
Okay I've updated changelog and version here so I think this and #267 can be merged and then released.

I don't have merge rights here just FYI so you'll have to do that thank you!

EdwinGuzman:
Published!

PR #268, opened 2020-07-13
add application/vnd.librarysimplified.web-epub to MediaType

M0nica:
We will need to use the `application/vnd.librarysimplified.web-epub` and `application/vnd.librarysimplified.axisnow+json` types in `circulation-patron-web` for Exploded EPubs.

### DIFF
--- a/packages/opds-web-client/src/interfaces.ts
+++ b/packages/opds-web-client/src/interfaces.ts
@@ -15,7 +15,8 @@ export type MediaType =
   | "application/audiobook+json"
   | "application/vnd.overdrive.circulation.api+json;profile=audiobook"
   | "application/vnd.overdrive.circulation.api+json;profile=ebook"
-  | "application/vnd.librarysimplified.web-epub";
+  | "application/vnd.librarysimplified.web-epub"
+  | "application/vnd.librarysimplified.axisnow+json";
 
 export interface MediaLink {
   url: string;
--- a/packages/opds-web-client/src/interfaces.ts
+++ b/packages/opds-web-client/src/interfaces.ts
@@ -15,7 +15,6 @@ export type MediaType =
   | "application/audiobook+json"
   | "application/vnd.overdrive.circulation.api+json;profile=audiobook"
   | "application/vnd.overdrive.circulation.api+json;profile=ebook"
-  | "application/vnd.librarysimplified.web-epub"
   | "application/vnd.librarysimplified.axisnow+json";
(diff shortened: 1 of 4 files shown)


packages/opds-web-client/src/interfaces.ts
```diff
@@ -14,7 +14,9 @@ export type MediaType =
   | "text/html;profile=http://librarysimplified.org/terms/profiles/streaming-media"
   | "application/audiobook+json"
   | "application/vnd.overdrive.circulation.api+json;profile=audiobook"
-  | "application/vnd.overdrive.circulation.api+json;profile=ebook";
+  | "application/vnd.overdrive.circulation.api+json;profile=ebook"
+  | "application/vnd.librarysimplified.web-epub"
+  | "application/vnd.librarysimplified.axisnow+json";
```
kristojorg, line 19: I'm not totally certain that all of these should be added here since my understanding is that the `.web-epub` type is never returned from the CM, but is only visible and used within CPW. Can you describe for me how you're planning to implement this or maybe we can schedule a conversation to do that?

packages/opds-web-client/src/utils/file.ts
```diff
@@ -13,6 +13,14 @@ export const typeMap: Record<MediaType, { extension: string; name: string }> = {
     extension: ".epub",
     name: "EPUB"
   },
+  "application/vnd.librarysimplified.web-epub": {
+    extension: ".xml",
+    name: "WEB EPUB"
```
kristojorg, line 18: I think as we add new items we should try to make these as "human friendly" as possible, since they will be displayed. Is this the way we would want to display these books to the user?

packages/opds-web-client/src/interfaces.ts
EdwinGuzman: Can you add me to the convo as well?

kristojorg: Yep will do!

M0nica: For the test server I am using with circulation-patron-web the type of the exploded EPubs is `application/vnd.librarysimplified.web-epub` however, now I know circulation manager doesn't return that type but returns `application/vnd.librarysimplified.axisnow+json`. I will extend the `MediaType` directly in circulation-patron-web to include `"application/vnd.librarysimplified.web-epub"` if that ends up being necessary.

PR #271, opened 2020-07-27
SAML Auth

kristojorg:
This PR adds the ability to handle SAML auth providers to opds-web-client. This also makes it possible to use SAML auth in circulation-patron-web. 

With SAML, the auth document now can contain one SAML provider that has many IdPs within it, each representing a sign in "option" for the end user. I chose to flatten these into a single array of auth providers containing both the SAML providers and any others configured.

### DIFF
--- a/packages/opds-web-client/src/utils/auth.ts
+++ b/packages/opds-web-client/src/utils/auth.ts
@@ -5,7 +5,7 @@ import {
   ServerSamlMethod
 } from "./../interfaces";
 
-export const SAML_AUTH_ID = "http://librarysimplified.org/authtype/SAML-2.0";
+export const SAML_AUTH_TYPE = "http://librarysimplified.org/authtype/SAML-2.0";
 
 export function generateCredentials(username: string, password: string) {
   const btoaStr = btoa(`${username}:${password}`);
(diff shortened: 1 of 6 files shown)


packages/opds-web-client/src/utils/auth.ts
```diff
@@ -1,4 +1,44 @@
...
+      type: provider.method.type,
+      description: getEnglishValue(idp.display_names) ?? "Unknown SAML Provider"
+    },
+    id: idp.href,
+    plugin: provider.plugin
+  }));
+}
+
+export const getEnglishValue = (arr: [{ language: string; value: string }]) =>
+  arr.find(item => item.language === "en")?.value;
```
EdwinGuzman, line 39: Eventually, this would be good to be something like `getLanguageValue` and the language "en" is passed in. But not needed right now. Do we imagine this to be a different language any time soon or ever?

kristojorg: I asked Leonard this and he said that he can envision sometime in the future that they might internationalize. I can change it if you like but I don't think that's going to happen anytime soon.

kristojorg:
Yea basically the way the `auth_document` structures SAML providers, there are multiple IdPs within one provider, which conceptually each map to a separate auth "option" for the user. With every other provider type, there is only one conceptual "option" per provider. So this function is only used to remedy that difference that is particular to how SAML providers are structured.

kristojorg:
FYI I'm not authorized to merge so if you don't mind doing that when you're ready that would be helpful, thank you!

packages/opds-web-client/src/utils/auth.ts
EdwinGuzman: Yeah, it was just something to consider later on.

EdwinGuzman:
Done.

PR #272, opened 2020-07-27
Fix url resolve

kristojorg:
We previously used the legacy `resolve` method from node's `url` module in the `OPDSDataAdapter`. This caused certain nested urls to be improperly resolved. This PR updates the adapter to use the new URL module, which properly resolves the urls. 

It has been tested in circulation-patron-web.


EdwinGuzman:
Cool, this is fine but can you update the version so I can publish a new version? Also,

kristojorg:
Yep, done along with changelog!

kristojorg:
Turns out I also forgot to update the changelog in 0.5.0 (#271 ). I added an item for it in this PR.

EdwinGuzman:
Published.

PR #273, opened 2020-07-27
fix read online label for atom links

kristojorg:
This PR is to make it so that links with indirect type `text/html;profile="http://librarysimplified.org/terms/profiles/streaming-media"` will show the "Read Online" download label instead of "Download atom", since they lead to the overdrive online viewer. 

I do a bit of refactoring of how streaming/indirect media is handled. I renamed "isStreaming" to "isReadOnline" because that seems more explicit. I also stopped exporting `isIndirect` from `useDownloadButton`, because it wasn't being used and it wasn't actually very accurate. This does mean CPW will need a corresponding PR to match the updating API of `useDownloadButton`.

### DIFF
--- a/packages/opds-web-client/CHANGELOG.md
+++ b/packages/opds-web-client/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## Changelog
 
+### v0.5.2
+
+- Fix: Show "Read Online" label for atom media types.
+
 ### v0.5.1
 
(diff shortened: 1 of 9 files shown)


kristojorg:
Yea it turns out those quotations are needed and that's why this wasn't working as intended to begin with. @EdwinGuzman or @M0nica if you have merge rights do you mind merging and publishing a new version when possible?

PR #275, opened 2020-07-31
add isPlatformCompatible

kristojorg:
This adds a flag to the return of `useDownloadButton` to indicate whether the file is platform compatible or not. Currently the only check we perform for this is whether the platform is mac and the file type is ACSM. James tells me there is no way to read these files on mac, and so they shouldn't be displayed.


kristojorg:
Closing this as I'm going to move the feature directly into CPW

PR #277, opened 2020-08-06
Adds support for multiple borrow links

hokei:
It's possible for a book to have multiple borrow links in an OPDS feed.  This exposes all of those URLs so that they can be used - I made it another entry in `BookData` so that it's not breaking other work, but it's not ideal.

### DIFF
--- a/packages/opds-web-client/src/OPDSDataAdapter.ts
+++ b/packages/opds-web-client/src/OPDSDataAdapter.ts
@@ -18,7 +18,8 @@ import {
   BookData,
   LinkData,
   FacetGroupData,
-  SearchData
+  SearchData,
+  FulfillmentLink
 } from "./interfaces";
 
 const resolve = (base, relative) => new URL(relative, base).toString();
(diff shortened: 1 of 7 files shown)


packages/opds-web-client/src/OPDSDataAdapter.ts
```diff
@@ -118,12 +119,26 @@ export function entryToBook(entry: OPDSEntry, feedUrl: string): BookData {
...
-    );
-  });
+  let allBorrowLinks: FulfillmentLink[] = entry.links
+    .filter(link => {
+      return (
+        link instanceof OPDSAcquisitionLink &&
+        link.rel === OPDSAcquisitionLink.BORROW_REL
+      );
+    })
+    .map(link => {
```
EdwinGuzman, line 129: This part is new. What's this doing?

hokei: Yeah 
https://github.com/NYPL-Simplified/opds-web-client/pull/277/files/7588148a1a90dea7f8b80cb9d35b24d9fb29b5e6..ed35a5b0224aca1313cf9d8d2cdbf47ed9ace736#diff-60c3f73af73e89b9faef5deab53edec4R63

So this wasn't actually typechecking before and the thing that was being returned was an object of type `OPDSAcquisitionLink` so what this is doing is returning a type of `FulfillmentLink` -- if you look at the FulfillmentLinks stuff before, this code is pretty much the same.  

I didn't extract it because I don't think that there's a reason these implementations should be tied together.  

But the idea is that all the info I need is in the `FulfillmentLink` type.  `OPDSAcquisitionLink` is a lot of extraneous information that just makes the `BookData` object harder to mock in CPW.

hokei:
@EdwinGuzman after merging should it be manually published to npm?

EdwinGuzman:
Yes, just released the new version `0.5.3`.

hokei:
thank you!

PR #280, opened 2020-08-28
Workaround to handle fulfillment redirects to S3

kristojorg:
This is a fix for a relatively complicated and annoying issue. Basically, when we attempt to download a book, we make an XHR GET request to the CM fulfillment link with our `Authorization` header. The CM then redirects us to wherever the content is ultimately hosted (sometimes through multiple redirects). In the case of DPLA Open Bookshelf, the CM was redirecting ultimately to an S3 bucket. The client would send that url the same request it originally sent to the CM, including the `Authorization` header. S3 did not like that header and would fail with a 400 response. 

To solve this, when a request for a blob fails, I check to see if the failure was after a redirect using `request.redirected`. If it was after a redirect, I retry the fetch without the `Authorization` header. 

### Some things to note:

- Doesn't work in IE (`request.redirected`) is not supported there. To solve this we could have the CM add a `X-Redirect-Url` or similar header in the original redirection. Or we could implement a non-302 redirect pattern, like sending a different status code with the redirect url in the body or another header. 
- The `Authorization` header is still being sent to anyone in the redirect chain. If we want that to stop, we need to implement the above described non-302 redirect pattern.
- I did try to "manually" handle redirects and strip the auth header before making the redirected request, but that is not possible due to browser security. You can either disallow redirects entirely, or follow them transparently. Nothing in between.
- This does introduce some additional latency as we have to wait for the original request to fully follow the redirect chain and then fail before we can detect it is a redirect and try again without the `Authorization` heaader.
- I don't think we control where content is ultimately hosted, so we may run in to more errors like this and need to strip out more headers or change our request in some other way in the future.

Interested to know if anyone else has suggestions here? I presume the mobile clients are following redirects manually and stripping the auth header when following them. It's a pity that isn't possible in the browser.

### DIFF
--- a/packages/opds-web-client/src/actions.ts
+++ b/packages/opds-web-client/src/actions.ts
@@ -87,7 +87,11 @@ export default class ActionCreator {
       dispatch(this.request(type, url));
       try {
         const response = await this.fetcher.fetch(url);
-        if (response.ok) return await response.blob();
+        if (response.ok) {
+          const blob = await response.blob();
+          dispatch(this.success(type));
+          return blob;
+        }
         /**
          * If this the response errored after a redirect, try again
          * without the Authorization header, as it causes errors when
@@ -99,18 +103,23 @@ export default class ActionCreator {
           });
           const blob = await newResp.blob();
           if (newResp.ok) {
+            dispatch(this.success(type));
             return blob;
           }
           console.error("Original Response ", response);
           console.error("New Response ", newResp);
-          throw new Error(`Retried fetch after redirect did not succeed.`);
+          throw {
+            status: 500,
+            response: `Retried fetch after redirect. The retry did not succeed.`,
+            url
+          };
         }
         console.error("Response: ", response);
-        throw new Error(
-          `Response was not okay and was not retried (wasn't the result of a redirect). Response:\n${JSON.stringify(
-            response
-          )}`
-        );
+        throw {
+          status: 500,
+          url,
+          response: `Response was not okay and was not retried (wasn't the result of a redirect).`
+        };
       } catch (e) {
         dispatch(this.failure(type, e));
         throw e;
--- a/packages/opds-web-client/src/actions.ts
+++ b/packages/opds-web-client/src/actions.ts
@@ -83,31 +83,38 @@ export default class ActionCreator {
   }
 
   fetchBlob(type: string, url: string) {
-    return (dispatch): Promise<Blob> => {
+    return async (dispatch): Promise<Blob> => {
       dispatch(this.request(type, url));
-      return new Promise<Blob>((resolve, reject) => {
-        this.fetcher
-          .fetch(url)
-          .then(response => {
-            if (response.ok) {
-              return response.blob();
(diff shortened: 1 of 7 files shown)


packages/opds-web-client/src/actions.ts
```diff
@@ -83,31 +83,47 @@ export default class ActionCreator {
   }
 
   fetchBlob(type: string, url: string) {
-    return (dispatch): Promise<Blob> => {
+    return async (dispatch): Promise<Blob> => {
```
EdwinGuzman, line 86: Do we want to throw errors rather than rejecting the error to display some kind of error to the user?

EdwinGuzman:
The code is fine but I'm not sure how to best handle the auth in the headers during the redirect - I don't know how bad it is to have it in the redirect chain. For the latency issue, maybe a loading bar can be introduced later to let users know something is happening but it's just taking longer.

packages/opds-web-client/src/actions.ts
kristojorg: I think this does throw an error does it not? Not totally sure I follow you

EdwinGuzman: Before, the `catch` would catch any error thrown and return a Promise `reject`. But now this is not resolving or rejecting any values, it's just throwing an error. So just concerned that if that's not caught, the app will crash since an error is thrown, whereas if you catch the error and reject a value, the app can continue working and a friendly error message can be rendered. I'm also not sure how an async promise works without a resolve/reject so maybe it's a lack of knowledge for that.

kristojorg:
@EdwinGuzman yea so the auth in the headers is bad from a security standpoint, but it will require a change to the CM in order to fix. We'll have to implement redirects some other way. I opened a ticket in Jira for that here: https://jira.nypl.org/browse/SIMPLY-3028

packages/opds-web-client/src/actions.ts
kristojorg: Yea I'm not totally sure either actually now that I think of it. I believe that it should be the same. Previously during failure:

1. Caught in `fetchBlob`, sent `failure` action
2. Rejected the promise. This had to be caught in a parent catch block, or else it results in an unhandled promise rejection error.

Now:

1. Dispatch the `failure` type
2. Rethrow the error. This has to similarly be caught by a parent catch block or else it causes an error.

So I believe it's the same... see here for more info: https://javascript.info/promise-error-handling#unhandled-rejections

kristojorg:
@EdwinGuzman can we get this merged when you have a chance?

EdwinGuzman (approved):

kristojorg:
@EdwinGuzman I'll resolve these conflicts but just so you know, I'm not approved to merge in this repo so you'll have to do that for me.

leonardr:
I filed https://jira.nypl.org/browse/SIMPLY-3078 to come up with a long-term solution to this problem, based on an earlier discussion we had.

EdwinGuzman:
@kristojorg deployed!

PR #282, opened 2020-08-31
type fixes

kristojorg:
This PR just has a few type fixes that will allow CPW to enable strict type checking.

### DIFF
--- a/packages/opds-web-client/src/AuthPlugin.ts
+++ b/packages/opds-web-client/src/AuthPlugin.ts
@@ -7,15 +7,15 @@ import { AuthMethod, AuthCredentials } from "./interfaces";
 /** Applications can implement this interface if they would like to support authentication
     methods other than basic auth. A list of `AuthPlugin`s can be passed as a prop to the
     `OPDSCatalog` component. */
-interface AuthPlugin {
+interface AuthPlugin<T extends AuthMethod = AuthMethod> {
   type: string;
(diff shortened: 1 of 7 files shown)


packages/opds-web-client/src/AuthPlugin.ts
```diff
@@ -7,15 +7,15 @@ import { AuthMethod, AuthCredentials } from "./interfaces";
 /** Applications can implement this interface if they would like to support authentication
     methods other than basic auth. A list of `AuthPlugin`s can be passed as a prop to the
     `OPDSCatalog` component. */
-interface AuthPlugin {
+interface AuthPlugin<T extends AuthMethod = AuthMethod> {
```
EdwinGuzman, line 10: Is the `= AuthMethod` similar to passing a default type if the caller just calls `AuthPlugin` without a type after?

EdwinGuzman (approved):

EdwinGuzman:
Can you just update the changelog/version for a new npm version?

packages/opds-web-client/src/AuthPlugin.ts
kristojorg: Yep exactly

kristojorg:
Okay did that!

kristojorg:
Could you merge this when you have a minute if you don't mind?

EdwinGuzman (approved):

